PR #3, opened 2012-11-19
added methods to open and access a submodule's repository

antis81:
further changed initialization and isValid() method

### DIFF
--- a/src/libGitWrap/Submodule.cpp
+++ b/src/libGitWrap/Submodule.cpp
@@ -222,8 +222,7 @@ namespace Git
 
     void Submodule::close()
     {
-        if (mMyRepo)
-            mMyRepo->deref();
+        mMyRepo = NULL;
     }
 
 }
--- a/src/libGitWrap/Submodule.cpp
+++ b/src/libGitWrap/Submodule.cpp
@@ -163,6 +163,23 @@ namespace Git
         return ObjectId::fromRaw( oid->id );
     }
 
+    ObjectId Submodule::indexOid() const
+    {
+        git_submodule* sm = getSM( mOwnerRepo, mName );
+        if( !sm )
+        {
+            return ObjectId();
+        }
+
+        const git_oid* oid = git_submodule_index_oid( sm );
+        if( !oid )
+        {
+            return ObjectId();
+        }
+
+        return ObjectId::fromRaw( oid->id );
+    }
+
     ObjectId Submodule::wdOid() const
     {
         git_submodule* sm = getSM( mOwnerRepo, mName );
--- a/src/libGitWrap/Submodule.cpp
+++ b/src/libGitWrap/Submodule.cpp
@@ -72,7 +72,7 @@ namespace Git
 
     bool Submodule::isValid()
     {
-        return ( mOwnerRepo && !mName.isEmpty() );
+        return ( mOwnerRepo && !mName.isEmpty() && getSM( mOwnerRepo, mName ) );
     }
 
     QString Submodule::name() const
--- a/src/libGitWrap/Submodule.cpp
+++ b/src/libGitWrap/Submodule.cpp
@@ -39,26 +39,29 @@ namespace Git
     }
 
     Submodule::Submodule()
-        : mRepo( NULL )
+        : mOwnerRepo( NULL )
+        , mMyRepo( NULL )
     {
     }
 
     Submodule::Submodule( Internal::RepositoryPrivate* repo, const QString& name )
-        : mRepo( repo )
+        : mOwnerRepo( repo )
+        , mMyRepo( NULL )
         , mName( name )
     {
-        isValid();	// do a lookup, so we know we're valid at least once
     }
 
     Submodule::Submodule( const Submodule& other )
+        : mOwnerRepo( other.mOwnerRepo )
+        , mMyRepo( other.mMyRepo )
+        , mName( other.mName )
     {
-        mRepo = other.mRepo;
-        mName = other.mName;
     }
 
     Submodule& Submodule::operator=( const Submodule& other )
     {
-        mRepo = other.mRepo;
+        mOwnerRepo = other.mOwnerRepo;
+        mMyRepo = other.mMyRepo;
         mName = other.mName;
         return *this;
     }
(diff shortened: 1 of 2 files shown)


src/libGitWrap/Submodule.cpp
```diff
@@ -172,4 +163,44 @@
...
+        {
+            return ObjectId();
+        }
+
+        return ObjectId::fromRaw( oid->id );
+    }
+
+    Repository Submodule::repository() const
+    {
+        return Repository( const_cast<Internal::RepositoryPrivate *>(*mMyRepo));
```
scunz, line 202: `*mMyRepo` is a nice construct here... Get's you think a bit, but should work nicely. I think, usually "mMyRepo.data()` is easier to understand; but we don't have that (yet) :smile:

src/libGitWrap/Submodule.h
```diff
@@ -72,10 +72,21 @@
...
+        ObjectId wdOid() const;
+
+    public:
+        Git::Repository repository() const;
+
+        /**
+         * @brief Opens a submodule's repository.
+         * @return true, when repoitory could be opened successfully; false otherwise
+         */
+        bool open(Result &result);
```
scunz, line 112: Could you add an `close()` here, too? It shouldn't do anything more than set `mMyRepo` to `NULL` again. Otherwise the new `RepoManager` won't be able to drop the reference count to zero and it will stay in memory (even if all of MGV thinks it was unloaded).

src/libGitWrap/Submodule.cpp
```diff
@@ -69,19 +72,7 @@
...
-
-        git_submodule* sm = getSM( mRepo, mName );
-        if( !sm )
-        {
-            mRepo = NULL;
-            mName = QString();
-        }
-
-        return mRepo;
+        return ( mOwnerRepo && !mName.isEmpty() );
```
scunz: There is a silent change here: `getSM` would have checked that `mName` is actually a submodule in the owning repository. This doesn't happen here anymore. Do we need it?

scunz:
All in all, this looks to me like a good thing to do. I'm just a bit curious about how we can safely force unloading (But as it looks, this is a thing I would have come accross anyway in the near future). Let's go with a `close()` for now; I will probably turn Submodule's data members into a ref-counted sub object, once this is really required.

src/libGitWrap/Submodule.cpp
antis81: Ooops, that's right. Currently we're not using the isValid() anywhere and I actually doubt we need it. Nevertheless, I'll correct ... It should not change anything though.

src/libGitWrap/Submodule.cpp
antis81, line 202: Thanks :smile:. Btw. this would crash when mMyRepo is NULL right ? :open_mouth:

scunz, line 202: No, it should not crash.
- `*myRepo` => will call to `const T* GitPtr<T>::operator *() const`, which just does `return d;`.
- The const cast actually does nothing.
- The `Repository` constructor creates an invalid Repository, if it's parameter is `NULL`.

Everything's fine :-)

src/libGitWrap/Submodule.cpp
```diff
@@ -172,4 +163,67 @@
...
+        if (!result)
+            return false;
+
+        mMyRepo = new Internal::RepositoryPrivate( submodule_repo );
+        return true;
+    }
+
+    void Submodule::close()
+    {
+        if (mMyRepo)
```
scunz: This is a bit dangerous. It dereferences without resetting `mMyRepo`.

`mMyRepo = NULL;` should do the trick.

src/libGitWrap/Submodule.cpp
scunz: It was used from the constructor and did set the name to an empty string, if the submodule wasn't valid. What i had in mind those days, was:

```
Git::Submodule s( mRepo, "Anything-that-does-not-exist" );
Q_ASSERT( !s.isValid() );
```

But that will work with the new code, too.

src/libGitWrap/Submodule.cpp
antis81: Thanks! Was unsure about that. I'll change it today evening, if you don't mind.

scunz: No need to hurry :-)

Just remember: Setting it to `NULL` is actually _all_ that is required (Think that wasn't clear from what I wrote above).

src/libGitWrap/Submodule.h
```diff
@@ -72,10 +77,48 @@
...
+
+        /**
+         * @brief Decreases the @see GitPtr of the submodule's repository.
```
scunz: We should not refer to internal classes in public API-Docs, should we?

antis81: Guess you're right. Somehow we need to express, what the method really does on the user side. Any idea?

scunz: Hm, it "closes the repository". Maybe adding that it does that soely by "dereferencing it, which might have the side effect that it isn't removed from memory immediately". :smile:

antis81: Thanks for the inspiration. Ok, after rethinking it "close" is the closest explaination - hm, sounds logical :smile:.

antis81:
ready for merge.

scunz:
```
[ 89%] Building CXX object src/Modules/Submodules/CMakeFiles/ModSubmodules.dir/SubmodulesCreateEditDlg.cpp.o
/var/lib/jenkins/home/w/MacGitver_GitSMI/src/src/Modules/Submodules/SubmodulesCreateEditDlg.cpp:41:30: error: no member named 'currentSHA1' in 'Git::Submodule'
    txtSHA1->setText( module.currentSHA1().toString() );
                      ~~~~~~ ^
1 error generated.
```

Seems this needs some fixes in MGV before it can be integrated.

antis81:
Yep, sorry I forgot to mention. The changed name is closer to lg2 and makes it more clear what the function does.

scunz:
> changed name is closer to lg2.

Then already: "Happy changing" for in about another week :-)

antis81:
OK, I'll keep that in mind :smile: - thanks.

scunz:
Just have a look at libgit2's PR numbered 1087. There's a giant change creeping in :-)

PR #6, opened 2013-01-11
Cloning

scunz:
Make clone work.


scunz:
bringing this change back to life: Remove the objective to implement fetching and checkout here; rather get the already working stuff in.

PR #7, opened 2013-01-12
An attempt for clearer docs and easier api usage

scunz:


scunz:
@antis81 Could you have a short look [here](https://github.com/macgitver/libGitWrap/blob/30749251189229103da50a7e3fe832b64cb4cc79/libGitWrap/Repository.hpp#L131-L173) and tell me, whether you think that a change like this might improve readability of the gitwrap-sources?

antis81:
Your changes seems quite readable to me. I'd change the method name `allRemotes` to `remotes` though. This goes good with `submodule` and `submodules`.

Not sure about the `diffSomethingToSomething` calls. While they are readable, my intuition says they should be either overloaded or in a `Differ` class, that belongs to the Repository. But that would mean a bigger structural change though.

scunz:
allRemotes vs submodules is a thing that I also stumbled upon. But it shouldn't return a QStringList either - and if it does, it should be named `remoteNames()` and we should provide a `remotes()` returning a `QList< Remote >`...

As for the diff-thingies... Well, there's actually a reason why i put them there (and even didn't consider to make their names shorter): Without those methods you need to get half a dozen temporary objects, which - including proper error handling - can fill a whole screen-page. Considering further that most of the time you need to call 2 of those diffs and combine their output together... I'd simply prefer them as a "short mess" than a logically structured nightmare... :smile:

so, should i go on[¹] and structure the gitwrap header files in this "doxygen-output layout"?

[¹] need to get the "fetching" branch done before that, anyway...

antis81:
Yeah, nothing against going on. Just change the `allRemotes` method-name to `remoteNames` then. Seems clear enough for me.

scunz:
I'll abandon this PR. By the time I'll get around finishing such a big cleanup, the work started here won't apply to the code base anyway. However, the concept is clear: Make it somewhat readable - that can be done at anytime...

PR #8, opened 2013-02-13
Improved DiffList

antis81:
The DiffList now uses a ChangeListEntry struct instead of single variables in the callback. not sure about the `QVector<ChangeListEntry> DiffList::changes()` method. I'll try that out in one of the next commits.

Going further: Thinking about the PatchConsumer, it uses the same variables in the startFile() callback. It could use the same struct here, right? If so, I would like to change it in this PR.

Btw.: There was a little bug in the changeListCallBack() method a wrong handled bool. That caused only the first change to be noticed and returned an error after that -> Fixed it.

### DIFF
--- a/libGitWrap/DiffList.cpp
+++ b/libGitWrap/DiffList.cpp
@@ -22,11 +22,35 @@
 #include "RepositoryPrivate.hpp"
 #include "DiffListPrivate.hpp"
 
+
 namespace Git
 {
 
     namespace Internal
     {
+
+        class DiffListConsumer : public ChangeListConsumer
+        {
+        public:
+            bool raw(const Git::ChangeListEntry &entry);
+
+            const Git::ChangesList &changes() const;
+
+        private:
+            Git::ChangesList     mChanges;
+        };
+
+        bool DiffListConsumer::raw(const Git::ChangeListEntry &entry)
+        {
+            mChanges.append( entry );
+
+            return true;
+        }
+
+        const ChangesList &DiffListConsumer::changes() const
+        {
+            return mChanges;
+        }
 
         DiffListPrivate::DiffListPrivate( const GitPtr< RepositoryPrivate >& repo,
                                           git_diff_list* difflist )
@@ -267,6 +291,14 @@ namespace Git
                                    consumer );
 
         return result;
+    }
+
+    ChangesList DiffList::changes(Result &result)
+    {
+        Internal::DiffListConsumer consumer;
+        consumeChangeList(&consumer, result);
+
+        return consumer.changes();
     }
 
 }
--- a/libGitWrap/DiffList.cpp
+++ b/libGitWrap/DiffList.cpp
@@ -34,10 +34,10 @@ namespace Git
         public:
             bool raw(const Git::ChangeListEntry &entry);
 
-            const Git::ChangesList &changes() const;
+            const Git::ChangeList &changeList() const;
 
         private:
-            Git::ChangesList     mChanges;
+            Git::ChangeList      mChanges;
         };
 
         bool DiffListConsumer::raw(const Git::ChangeListEntry &entry)
@@ -47,7 +47,7 @@ namespace Git
             return true;
         }
 
-        const ChangesList &DiffListConsumer::changes() const
+        const ChangeList &DiffListConsumer::changeList() const
         {
             return mChanges;
         }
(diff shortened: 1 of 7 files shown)


scunz:
I agree, that `PatchConsumer::startFile` is the same callback as issued from `ChangeListConsumer`.
But thinking further about this, isn't it actually like this: A `PatchConsumer` is a more detailed version of a `ChangeListConsumer`, which would translate to C++ as `PatchConsumer` should inherit `ChangeListConsumer`?

antis81:
Yeah, inheritance seems to be really a good option. I added the rest of the changes to be able to call DiffList::changes(). Are you ok with that?

libGitWrap/DiffList.hpp
```diff
@@ -24,13 +24,17 @@
 
     class Repository;
     class ChangeListConsumer;
+    struct ChangeListEntry;
     class PatchConsumer;
 
     namespace Internal
     {
         class DiffListPrivate;
+        class DiffListConsumer;
```
scunz: actually not needed as forward decl, is it?

libGitWrap/DiffList.hpp
```diff
@@ -24,13 +24,17 @@
...
     }
 
+    typedef QVector<ChangeListEntry> ChangesList;
```
scunz: Shouldn't it be singular? i.e. `ChangeList`?

libGitWrap/DiffList.cpp
```diff
@@ -264,4 +293,12 @@
         return result;
     }
 
+    ChangesList DiffList::changes(Result &result)
```
scunz: Thinking again about my own naming: would be more consistent if we named it `changeList()` (like the result code)

scunz:
Yepp, the general change is nice this way. (Just a few nitpicks amended).

libGitWrap/DiffList.hpp
antis81: Yes, this one is obsolete. I'll remove it.

libGitWrap/DiffList.hpp
antis81: Probably would be better in singular.

libGitWrap/DiffList.cpp
antis81: Ok :smile:.

antis81:
Another review, please. I implemented the inherited version of the PatchConsumer (it changes libMacGitverCore as well).

antis81:
:question: @scunz Unfortunately, there's documentation missing on lg2 about the `git_index_entry` struct. Maybe you have a better idea what the members mean:
- git_index_time ctime
- git_index_time mtime
  Is it creation and modified date?
- unsigned int dev
- unsigned int ino
- unsigned int mode
- unsigned int uid
- unsigned int gid
- unsigned short flags
- unsigned short flags_extended

Clear to me are these:
- char *path
- git_off_t file_size
- git_oid oid

The current (test)implementation simply runs through the index with a for loop and outputs the current entry's path. It actually shows all files in HEAD, wether they are staged or not.

# Edit: Moved this discussion to PR #10.

libGitWrap/IndexEntry.cpp
```diff
@@ -0,0 +1,125 @@
...
+QTime IndexEntry::cTime() const
+{
+    git_index_time t = d->mEntry->ctime;
+    return QTime( 0, 0, t.seconds );
+}
+
+QTime IndexEntry::mTime() const
+{
+    git_index_time t = d->mEntry->mtime;
+    return QTime(0, 0, t.seconds );
```
scunz: Just a gut feeling without deeper research: isn't [QDateTime::setTime_t](http://doc.qt.digia.com/stable/qdatetime.html#setTime_t) the correct function to convert these?

scunz:
@antis81, could you separate this into 2 topics: One for the ChangeList changes and one for the Index.

I think, the ChangeList-Thingie looks more or less finished and could be merged; but related to the IndexEntry additions, I see a big issue with object life times, which I'd rather like to address without the ChangeList changes being intermingled into the diff views... :-)

libGitWrap/IndexEntry.cpp
antis81: Well, the main problem here is, that git_time is a uint64, but QDateTime and time_t use uint fields. This might be wrong for some 32-bit systems :bomb:.

antis81:
Already thought about splitting branches. So, if you don't mind I would move the `topic/ImprovedDiffList` branch (edit: two or three commits before) and set up a new one based on current `development` (edit: for the index).

libGitWrap/IndexEntry.cpp
scunz: Hm, but the resolution is _seconds_. Meaning it will wrap around the 32bit border 136 _YEARS_ after 1.1.1970. Am I miscalculating?

scunz:
Yes, that sounds like a plan :-)

antis81:
Done.

scunz:
Cool, so you might now merge this one here (and it's associates) and be done with it :-)

PR #10, opened 2013-02-24
Git Index implementation

antis81:
PR to implement the git index functionality and merge it back to development.


libGitWrap/IndexEntry.cpp
```diff
@@ -0,0 +1,125 @@
...
+        , mUnmanaged( unmanaged )
+    {
+        Q_ASSERT( mEntry );
+    }
+
+    IndexEntryPrivate::~IndexEntryPrivate()
+    {
+        if( !mUnmanaged )
+        {
+            //git_index_entry_free( const_cast< git_index_entry* >( mEntry ) );
```
scunz: I assume you left this commented out because there is no `git_index_entry_free`-thing inside libgit2?

libGitWrap/IndexEntryPrivate.hpp
```diff
@@ -0,0 +1,48 @@
...
+{
+    namespace Internal
+    {
+
+        /**
+         * @internal
+         * @ingroup     GitWrap
+         * @brief       The IndexEntryPrivate class
+         */
+        class IndexEntryPrivate : public BasicObject
```
scunz, line 34: First off, in my book, implementing `IndexEntryPrivate` as `BasicObject` is the right way to go.

However, this will cause us major pain, if we keep storing a `const git_index_entry*` inside (see following comments).

libGitWrap/IndexEntry.cpp
antis81: Yes, that's it.

antis81:
Moved this question. Please do not answer:

:question: @scunz Unfortunately, there's documentation missing on lg2 about the `git_index_entry` struct. Maybe you have a better idea what the members mean:
- git_index_time ctime
- git_index_time mtime
  Is it creation and modified date?
- unsigned int dev
- unsigned int ino
- unsigned int mode
- unsigned int uid
- unsigned int gid
- unsigned short flags
- unsigned short flags_extended

Clear to me are these:
- char *path
- git_off_t file_size
- git_oid oid

The current (test)implementation simply runs through the index with a for loop and outputs the current entry's path. It actually shows all files in HEAD, wether they are staged or not.

antis81:
Reply from scunz:

Most of these fields are directly copied from output of [lstat(2)](http://linux.die.net/man/2/lstat).

dev = device id; ino = inode number in the file system; mode = file mode; uid = user id, gid = group id; ctime and mtime = creation / modification time.

As for the flags, i must take a look first => [see here](https://github.com/git/git/blob/master/Documentation/technical/index-format.txt#L85-105)

But I'm missing the "stage" field in your list, which is probably the most interesting one: It contains the "stage" which the index entry belongs to. As I see it, it's encoded into the flags.

scunz:
LibGit2 assumes to have full control over the index entries. The user may modify their contents (except the path name) but is never allowed to free, reallocate or create new index entries. (As you already mentioned, the index documentation in libgit2 seems to be a "bit behind"... However, one reads the `git2/index.h` file carefully, it is not hidden, that libgit2 wants to manage memory on its own here).

When lg2 gives us a `git_index_entry*` we cannot know how long that one will live. On one hand, the index could go totally out of scope; and on the other hand, a call to one of `git_index_remove`, `git_index_remove_directory` or `git_index_add` might free the entry.

We probably don't want to lookup the index entry for each and every query to the `Git::IndexEntry` either. So I see no other possibility than creating a copy of all data members of `git_index_entry` in `Git::Internal::IndexEntryPrivate`. That isn't nice, but the only way to play safe.

antis81:
@scunz Ok, I'll try to sum it up a little:
- I would revert to the first approach with a smart pointer holding a pointer to the original `git_index_entry`. That means: no internal copy(!)
- For completeness, I'll commit the already finished "smart-pointer-with-copy-solution" as well.
- Talk to the lg2 people (that's you also :smile:), that we need to know when the object is free'd, so we can clean up our smart pointer without ever touching the original. That goes exactly fine with what libgit2 wants - keep the index safe from unwanted changes :lock:.

scunz:
Hey @antis81, when I saw the comment last night, I thought: damn it that's not the real deal; We can't do it like that.
But now, that I'm looking at the code. Well, the version that is in this PR _right now_, **is quite cool**. There's two minor issues, but we'll get them sorted out for sure:
- `qstrdup` is deprecated in Qt5 and not exposed under all circumstances. Let's just use `strdup` :-)
- `free( mEntry->path );` is missing on the `~IndexEntryPrivate` destructor.

From a memory-allocation point of view, this is exactly as efficient as the `malloc`-based version that I proposed; It's however more C++'ish, so let's settle with this one.

scunz:
@antis81, in my ref-counting investigations, I made a commit that adds more construction options to `Git::Index`. I hope you don't mind that I cherry-picked that one onto this PR.

P.S.: The `free()` call is still missing, isn't it?

antis81:
Yeah, thanks. No I don't mind the cherry pick.

> P.S.: The free() call is still missing, isn't it?

Right, the free() and strdup() is still missing. I added this, but didn't commit yet, cause I was unsure if it works well with the static instance of `git_index_entry`. Do we need a Pointer instead or is it all fine?

scunz:
I think, this is looking fine now. Are we ready to merge?

antis81:
Yes, please :+1:.

scunz:
So let's let it in.

PR #12, opened 2013-03-19
Topic/git status

antis81:
Here's an attempt to implement the git_status. This will also change the calls in the worktree module in macgitver/libMacGitverModules (seperate PR follows).
- [x] implementation of the repo status
- [x] implementation of the submodule status

@scunz: `git_status_foreach` and `git_status_foreach_ex` are deprecated. Any idea how to get the files different.


scunz:
First of all: You have copied a bad mistake of mine, that i just spotted reading this PR. So, I should probably say thanks for that. I will document and fix it, then let you review, so you can adjust this code, too.

Who says that the status Api is deprecated? I know that it is insanely slow, but I've not seen a better thing yet. Unless, of course, you want to get the data directly from a diff (status is actually just iterating a difflist).

I'm not sure, a special class is required for status - but i don't mind it either.

antis81:
Hm, well seems like I fooled myself again. The git_status_foreach methods are marked red in the lg2 API docs for HEAD. And that should mark a remove, or not? But looking into development it is still exported in the public api. So, don't mind about it further for now.

scunz:
I honestly believe that this docurium-thingie is not worth looking at it at all. You simply can't put the 20 years of knowledge that Dimitri (doxygen) has into 500 lines of ruby...

The source code of libgit2 has a very good and detailed documentation level. Docurium perfectly manages to hide that fact.

scunz:
> First of all: You have copied a bad mistake of mine, that i just spotted reading this PR. So, I should probably say thanks for that. I will document and fix it, then let you review, so you can adjust this code, too.

I think that I'm starting to see pink :elephant:. I had the pattern in mind that I originally used; forgetting that I did a longer debugging session which resulted in the current `~IndexPrivate` destructor, which makes everything perfectly fine (and you actually used the same pattern; which means that your code is also perfectly fine).

Such things wouldn't happen, if we actually did test the code we write here. I assume 75% of the libGitWrap code is used _nowhere_.

antis81:
> I'm not sure, a special class is required for status - but i don't mind it either.

That's actually a good question. I actually tried to go with static calls and even thought about a singleton approach first, which basically seems like a good idea. But ... there's some advantages for using the class.
- At present, transporting an instance of `Git::Repository` to a static call causes a lot of headaches and would need further effort in refactoring to implement a mechanism.
- The class approach slims the Repository class in structure and makes it easier to handle.
- GitPtr calls are **threadsafe**. So accessing status information is too.
- There's one `Status` instance per repo => few overhead.

As always, I'm open for better solution :smile: :+1:.

scunz:
First to the minor issues:

> GitPtr calls are threadsafe. So accessing status information is too.

I believe it would be a bit foolish to reduce thread safety to the access of the smart pointer. i.e. One thread could easily call `git_index_read` on the repository's index while another one is running through `git_status_foreach` - I'm almost 100% sure this will bring a crashing race...

> The class approach slims the Repository class in structure and makes it easier to handle.

Actually the smallness of our files esp. in `libGitWrap` is a disgrace. I know that this is a contradiction to what I usually do. I've been developing for 4 years under the pressure of a coding rule that stated: "Declare/Implement no more than _one_ class per file" - I indeed know the benefits and the drawbacks of that. No sensible developer would follow that rule strictly. It's biggest drawback is the increase of compile times by magnitudes (It takes ~4 minutes to compile MGV where actually 1 might be enough with a less stricter file layout). However, a big advantage is that you're easily able to figure where a given piece of code is located.
I don't think we need to reduce our classes (_yet_).

> I actually tried to go with static calls and even thought about a singleton approach first, which basically seems like a good idea.

I don't see a reason for anything static here. Maybe that's either another misunderstanding or not cleanly thought through. What exactly would you want to solve with a static call or a singleton?

So, let me put the old status quo into a "verbal uml":
- There's one entity "Repository"
- There's one entity "StatusHash"
- The entity Repository has a method that produces a StatusHash-Entity for exactly this Repository-Entity

With your change it becomes:
- There's one entity "Repository"
- There's one entity "Status", which has a 1:1 relation to the entity "Repository" (and cannot be used standalone)
- There's one entity "StatusHash"
- The Repository-Entity has a method that returns the Status-Entity related to this Repository-Entity.
- The Status-Entity has a method that produces a StatusHash-Entity for the Repository-Entity to which this Status-Entity is related

The only static approach (a singleton actually) that i could imagine:
- There's one entity "Repository"
- There's a singleton entity "Status"
- There's one entity "StatusHash"
- The status entity has a method into which one inputs a Repository-Entity and it outputs a Status-Hash-Entity for the Repository-Entity

How should that be justified in a user-manual?

antis81:
> Actually the smallness of our files esp. in libGitWrap is a disgrace. ...

This is surely a hot topic :fire: and can easily lead to flame-war. But I'm totally with you and not a fan of such "strict rules". Let's discuss that in issue #13 (too early for a PR).

Back to topic:

> The only static approach (a singleton actually) that i could imagine:
> - There's one entity "Repository"
> - There's a singleton entity "Status"
> - There's one entity "StatusHash"
> - The status entity has a method into which one inputs a Repository-Entity and it outputs a Status-Hash-Entity for the Repository-Entity

This is roughly what I'm talking about. I took another look into code and found, that I missed the `GitPtr::data()`method. This is enough to achieve the Singleton approach :heart:.

Concerning thread safety you're probably right, but we can care about that at a later point in time.

scunz:
What exactly is the benefit of `hash = Git::Status::files( repo );` over `hash = repo.files()`?

antis81:
Phew, turns out that this is getting a bigger topic than I initially thought it would. The Repository class starts to get bloated ... and that's why I tried directly to put it separate. Especially having the submodule's status in mind. But I have to admit it is the wrong place to do it here.

It is better to call the file status for the repository and the submodule status for the submodule separately. Something like:
- `repo.status(const QString &relativePath);` // retrieves status for single file
- `submodule.status();` // retrieves submodule's status

scunz:
> The Repository class starts to get bloated.

That is actually a bullet point, I would completely agree to.

However:
A repository is a repository is a repository is a repository. Any repository inside it is still a repository < /sarcasm>

What I actually want to say is: We have a very thin and effective object model (predicted by the concepts of git). Why shouldn't we stick to that?
1. You open a repository via `Git::Repository::open()`. => `repo.status()` gives you the status of its files.
2. You have an opened repository and it has a submodule. There are different ways to access it; one being `repo.submodule( name )`, which returns a `Git::Submodule` where you can call to `repository()` and then you're absolutely able to do with the submodule any operation that would be possible in 1.

We could still shorten that with a signature like:
    `Git::Repository Git::Repository::subRepository( const QString& path )`
for example.

scunz:
I am afk till roughly midnight, but i think that i will setup an ER diagram for libGitWrap. Might be that this would be a better basis for discussions where which functions belong.

libGitWrap/GitWrap.hpp
```diff
@@ -91,27 +91,29 @@
...
-        StatusWorkingTreeModified   = (1 << 8),
-        StatusWorkingTreeDeleted    = (1 << 9),
-        StatusWorkingTreeTypeChange = (1 << 10),
+        WorkingTreeNew        = (1u << 7),
+        WorkingTreeModified   = (1u << 8),
+        WorkingTreeDeleted    = (1u << 9),
+        WorkingTreeTypeChange = (1u << 10),
 
-        StatusIgnored               = (1 << 14)
+        Ignored               = (1u << 14)
```
scunz: @antis81, I really like the idea of shortening those enumerator names. But I'm a bit concerned that `Ignored` and `Unchanged` might clash with some other things we might come up with in the future.

I see 3 ways around that: a) move the enum into the `Repository` class b) Keep those two values un-shortened c) anything you could come up with.

Which would you like more?

libGitWrap/Repository.cpp
```diff
@@ -69,6 +68,17 @@
...
         }
 
+        static int statusHashCB( const char* fn, unsigned int status, void* rawSH )
+        {
+            #if 0
+            qDebug( "%s - %s", qPrintable( QString::number( status, 2 ) ), fn );
+            #endif
+
+            Git::FileStatusHash* sh = (FileStatusHash*) rawSH;
+            sh->insert( QString::fromUtf8( fn ), FileStatusFlags( status ) );
```
scunz: When you're touching this code anyway: could you provide a wrapper method in `GitWrapPrivate.h` that turns the status code from libgit2 into one of ours - as opposed to just cast it around.

libGitWrap/Submodule.cpp
```diff
@@ -238,6 +238,19 @@
...
 
+    Submodule::Status Submodule::status(Result &result) const
+    {
+        if ( !result || !d )
+        {
+            return InHead;
+        }
+
+        unsigned int status = GIT_SUBMODULE_STATUS_IN_HEAD;
+        result = git_submodule_status( &status, d->getSM( result ) );
```
scunz, line 282: ah, now I'm actually getting what you were meaning with "submodule status". I wasn't aware of the `git_submodule_status` method.

libGitWrap/Submodule.hpp
```diff
@@ -55,6 +55,28 @@
             Ignore
         };
 
+        /**
+         * @brief The Status enum describes a submodule's status.
+         */
+        enum Status
+        {
+            InHead          = (1u << 0),
```
scunz: Could you add some `@var`-Doxygen tags for each enumerator value, please? I do know the lg2 code deeply, but I'm still confused as to what the actual meaning of these names is. I can only imagine a newbee to this code would be feeling totally lost about these names.

libGitWrap/GitWrap.hpp
antis81: The a) solution looks most appealing to me. We could use a `Status` namespace, but that is the same like the prefix. It already clashed btw. with the `Submodule::Status` - where I put it inside the class :innocent:.

libGitWrap/Repository.cpp
antis81: Sounds cool - I'll do that.

libGitWrap/Submodule.hpp
antis81: Ok.

libGitWrap/GitWrap.hpp
antis81: Ahem, well ... putting the enum in the class does not work well. This actually makes it worse:
1. Moving the enum to `GitWrapPrivate.h` (see comments below) becomes impossible (needs to include `Repository.h`, which we don't want).
2. We'd have to use e.g. `Git::Repository::IndexNew` from anywhere outside the `Repository` class - sigh :frowning:. Leads us to solution b) or - :bulb: - to introduce a `Git::File` namespace around the `FileStatus...` constructs. From the usage view this would go great, if you look at the whole thing. Would that be ok?

scunz:
btw.: Did you notice that PRs for GitWrap are CI-Checked again, too? :-)

antis81:
> Don't you think that the submodule-status and the file-status are kind of overlapping? I wonder if we could manage to keep it as enum Status and a typedef QFlags< Status > Stati, if we manage to map the submodule status codes to file status codes. What would you think about that?

Yeah, I like the idea of combining the stati. Not sure though, if it is possible at all to do it safely. The flags differ a lot. Maybe we should leave it separate for now.

scunz:
> Yeah, I like the idea of combining the stati. Not sure though, if it is possible at all to do it safely. The flags differ a lot.

I didn't have that impression, but you're probably right anyway: Let's go with something working and make optimizing decisions later.

antis81:
Hope the following dox commit will make it more clear. Information is taken from libgit2. The submodule status is actually slightly different in meaning from the file status (well, not a lot though :smile_cat:). Well ... we could convert it to a file status in addition, but currently I cannot imagine a use case for that.

scunz:
I still think that the sm-status is a superset of the file-status - and frankly from the status point of view i dont see a difference between a file/symlink and a submodule.

antis81:
Absolutely. But how to derive an enum? We'd need a "bigger" structure like a class or something around it.

I can think of the following:
1. Strip the sm-status and remove the IndexAdded ..., WorkingTreeAdded ... flags.
2. Setup a QMap Submodule member or something similar containing the two enums like:
key = "EntryStatus" -> v = the File::Status
key = "SubmoduleStatus" -> v = the stripped Submodule::Status
3. Split the git_submodule_status and map it using the just added "convertSubmoduleStatus" method.

Unclear: What happens to the two "GIT_STATUS_TYPE..." flags? They are missing in the information above.

So ... thinking about that it makes some sense. But I think we should leave it for later, when we really need such a feature, before putting the effort into for nothing. Maybe we should make up another issue in MGV 0.2 for that purpose, or?

scunz:
As per definition from my roadmap-thoughts, we could change all of this until a MacGitver **V1.0** release. But the sooner we aim for a stable API the better.
That being said, I don't see a reason why we cannot simply add all entries into one `Git::Status` enum and use the relevant flags only if they apply (i.e. never set submodule flags when we do a query for file status).

At some point we will be implementing icons in the WT/Index-View, if we go like this, we could just do `fs = getFileStatus(); if( isSubmodule ) fs |= getSubModuleStatus(); findIconFor( fs);`. If we keep them separated in any way, the findIconFor function would have to switch-case both enums or at least 2 times the same enum _plus_ the individual submodule state.

Please don't get me wrong, I'm _not_ trying to pick this apart as good as I can. But good API-design is a really complex business. I.e. most people don't even realize _why_ developing against Qt feels way much more comfortable than developing against MFC. The trick here is, that the total sum of such micro-decisions makes the difference. Like: What constructors are available? What parameters should they take? Where does an enum go? What scope does it have? Can I use code against class Foo unchanged but use class Bar instead, even if the have nothing else in common?

---

But back to the heart of the matter:

Actually, I don't understand why the submodule status is of so much importance to you in the context of what you're trying to achieve. Doesn't the file status give already enough information to implement viewing the state of Index and Work-Tree in one view? This isn't rhetorical, I'm really wondering what you want to do with that status.

Here's some thoughts:
- We should not try to present a combined tree of files inside a superproject and its submodules. Doing so, will probably bring a lot of pain: We need to implement interactions on individual WT-/Index-Entries (commit, checkout, reset etc.) and when the tree becomes a tree of trees, there's many corner cases to catch up with. Also we'd have to fiddle a lot with the result from diff.
- We don't have any icons yet, we can't even visually mark a folder as being a submodule. This is a todo item that is by itself very big and complex. The fact that we need to have the icons is only a minor part of it. We also need to adjust lots of code paths to actually use icons.
- We actually could show the submodule entry inside the WT/Index and decorate it specially, but:
- I don't even see a cheap way of determining per path whether that path is a submodule at all.

antis81:
:+1::+1::+1:. This answers exactly what I wanted to find out and I understand you totally. It is a lot of "nitpicking" when it comes to deciding for the "right" API. Most programmers care more about the "done" than about the "how" in "how it should be done".

So ... I'll do it and merge the two stati into one bigger `Status` enum.

Btw.: Currently, I don't need or use the submodule status, but I want to have it for completeness.

scunz:
> Most programmers care more about the "done" than about the "how" in "how it should be done".

Not to forget: the more of the stereotype "Manager" you merge into that stereotype "Programmer", the more the "how" vanishes and the "done" takes over. At a certain point, the "how" is completely gone. I don't dare to count up, how many times I had discussions like:

Manager: Why the hell do you think this simple-looking change takes 2 months to implement?
Me: Because _your_ decisions from last March, April and September ruled down the suggestions that _I_ made with respect to usability, maintainability and readability.

Gladly, we're not too complicated, yet. Did you ever realize that it can take up to 100 mails before you get a 10 line change into git.git or the linux kernel? 

> Currently, I don't need or use the submodule status, but I want to have it for completeness.

This is the best of all explanations :-)

scunz:
The https://github.com/macgitver/libGitWrap/commit/7921e62781dd4b85fcdfd6b9759414d57e0aab59 is a really nice change. Maybe we could rename the enum back to just `Status`?

antis81:
Strange, your anecdote reminds me of something - oh yeah, today morning :grinning:. Wow, no I didn't know that linux thingy. Well, good decisions take their time.

> Maybe we could rename the enum back to just Status?

Uhm, you mean removing the namespace? Of course we can. But I start to really like it :heart: :smile:. Don't you still think, it could clash easily with other names? OTOH ... when it really clashes, we can still find another solution.

scunz:
Yes, removing the namespace was what I was thinking about. Actually, the Unchanged and Ignored might clash some day. For clashing, Ingored is bothering me more, because it is likely to clash. We could go this way:

```
FileIgnored = (1u << 1234),
SubmoduleIgnored = FileIgnored,
```

We could probably do the same for Unchanged. If we'd even go so far, we could do that for _all_ values in the enum and then we're again pretty close to what you wanted in the first place.

But when it comes to Unchanged, I'm more worried that it is in theory possible that a file is tracked but included in `.gitingore` (libgit2.git has such files), then it might as well be Ignored and Unchanged at the same time, but `Unchanged | Ignored == Ignored` and I guess that libgit2 is already reporting it that way.

antis81:
Are we ready to merge this?

scunz:
Indeed, this is looking good from my POV. I'm not exactly sure how you feel about the last set of changes, butt I think they were good for consistency.

If you feel like merging this one now, just go ahead.

antis81:
Especially for the last commit, I had no idea how to split into several commits. Apart of that, I think this one is ready and I feel good to go on.

PR #14, opened 2013-04-03
Required enhancement for Git::Status

antis81:
Small followup for PR #12.


scunz:
This is looking good to me.

antis81:
Good, I'll merge then.

scunz:
Not sure if this should have  waited for the consumer code to change along. But at least there is no compiler errors...

antis81:
Yeah, I new that before and it is a very small non-destructive change, that's why :smile:.

PR #16, opened 2013-05-10
feature: reset/unstage index entries

antis81:
This one brings in the ability to unstage files from the index.

### DIFF
--- a/libGitWrap/GitWrap.cpp
+++ b/libGitWrap/GitWrap.cpp
@@ -58,10 +58,10 @@ namespace Git
         {
             internalCopy = sl;
 
-            a.count = sl.count();
+            a.count = internalCopy.count();
             a.strings = new char *[a.count];
 
-            for(int i=0; i < sl.count(); ++i )
+            for(int i=0; i < internalCopy.count(); ++i )
             {
                 a.strings[i] = internalCopy[i].toUtf8().data();
             }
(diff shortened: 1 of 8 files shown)


libGitWrap/GitWrap.cpp
```diff
@@ -46,6 +46,42 @@
...
+        }
+
+        StrArrayWrapper::StrArrayWrapper(const QStringList& sl)
+        {
+            internalCopy = sl;
+
+            a.count = sl.count();
+            a.strings = new char *[a.count];
+
+            for(int i=0; i < sl.count(); ++i )
```
scunz: should use internalCopy, too.

antis81:
Think we can merge this as well, or?

scunz:
Isn't the StrArrayWrapper (introduced here) used by the PR you just merged? I'm a bit confused. But anyway, this PR doesn't look bad either. So I'd say: go ahead.

antis81:
Yes. The `checkout` PR goes on this branch and not on `development`. It would have been better to keep that StrArray thingy in another `c2cpp_internal` branch (and PR respectively), that contains internal core functionality to multiple API methods ... Anyways, next time I'll do better :innocent:.

So I'm going to merge here and the `WorkingTree` feature branch and refactoring branches (which will need some cleanup in this branch for the "new" functions) in turn.

PR #17, opened 2013-05-11
feature: checkout commands

antis81:
TODO's:
- [x] Checkout a commit; equivalent to `git checkout <ref_or_sha1>`
- [x] Checkout files to index (discard unstaged changes)
- [x] Checkout files to specific revision; equivalent to: `git checkout <ref_or_sha1> -- file.txt`

### DIFF
--- a/libGitWrap/ObjectCommit.hpp
+++ b/libGitWrap/ObjectCommit.hpp
@@ -78,6 +78,7 @@ namespace Git
         QString message( Result& result GITWRAP_DEFAULT_TLSRESULT ) const;
         QString shortMessage( Result& result GITWRAP_DEFAULT_TLSRESULT ) const;
 
+        void checkout( Result &result GITWRAP_DEFAULT_TLSRESULT ) const;
         Reference createBranch( const QString& name, bool force,
                                 Result& result GITWRAP_DEFAULT_TLSRESULT ) const;
 
--- a/libGitWrap/ObjectCommit.hpp
+++ b/libGitWrap/ObjectCommit.hpp
@@ -77,7 +77,7 @@ namespace Git
         QString message( Result& result GITWRAP_DEFAULT_TLSRESULT ) const;
         QString shortMessage( Result& result GITWRAP_DEFAULT_TLSRESULT ) const;
 
-        void checkout( Result &result GITWRAP_DEFAULT_TLSRESULT,
+        void checkout( Result &result,
                        bool force = false,
                        const QStringList &paths = QStringList() ) const;
         Reference createBranch( const QString& name, bool force,
--- a/libGitWrap/ObjectCommit.hpp
+++ b/libGitWrap/ObjectCommit.hpp
@@ -17,8 +17,7 @@
 #ifndef GIT_OBJECT_COMMIT_H
 #define GIT_OBJECT_COMMIT_H
 
-#include <QDebug>
-#include <QList>
+#include <QStringList>
 
 #include "GitWrap.hpp"
 #include "ObjectId.hpp"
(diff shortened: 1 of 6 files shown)


antis81:
Ok, we are almost ready to merge this. I'd just like to do small finalization corrections to that. The two methods `Reference::checkout` and `ObjectCommit::checkout` are almost the same and differ just in one point. Reference updates the HEAD to point to the ref while a commit is always detached. Maybe we could merge these two somehow - or should we leave it like it is? Not sure.

libGitWrap/ObjectCommit.hpp
```diff
@@ -78,6 +77,9 @@
         QString message( Result& result GITWRAP_DEFAULT_TLSRESULT ) const;
         QString shortMessage( Result& result GITWRAP_DEFAULT_TLSRESULT ) const;
 
+        void checkout( Result &result GITWRAP_DEFAULT_TLSRESULT,
```
scunz: Putting the result parameter in the pole position deviates from actually everything. However, after only thinking for 20 seconds about this: I like it more. What should we do?

scunz:
For the update-the-HEAD part: I'm really not sure that the consumer of such API is always expecting the HEAD to be updated to anything at all.

Esp. in the context of Cookery, I wanted to do some _really_ crazy things: The design is to have a bare repository which contains a very specialized hierarchy of refs. Each such ref-subtree is a mirror of the ref-tree got from one remote, which gets updated (=fetch) on a regular (time/event/any trigger)-basis. This already helps me in getting rid of all the backup-update shit I'm currently doing...
If cookery then thinks any actions are due, for each action a separate checkout is done (if required by that action). This is i.e. true for building PRs. I'm pretty sure that I don't want to update the bare repo's HEAD (which equals to its default branch) when building a PR.

Does this make any sense in the context of a generalized GitWrap library?

libGitWrap/ObjectCommit.hpp
antis81: This was done on purpose due to the default parameters used here. Soo, if you don't mind I'd like to keep it like that. As it is always a must parameter it makes sense to keep it in pole position.

scunz: Indeed. However, I'm more worried about consistency; thus my question about what to do: change it everywhere or allow the deviation to slightly pass in?

antis81: Hm, my first thought was: "Ah well, let it slightly pass in." But you're right. We should really care about the API consistency and I also see that. Well, seems we both agree, it is a good solution to put the result param in front wherever that is possible. As this means bigger changes to the API, we better put it in a feature branch, right? I'd start work on that tonight right away from the current development branch and go through it quickly. Ok?

antis81:
> Does this make any sense in the context of a generalized GitWrap library?

Yeah, of course it makes sense and I'm glad you asked that. I just focused on MGV's use cases, like we said until today. If we want to move GitWrap in a more generic direction, we need to abstract calls a little better, so that they reflect both - the `plumbing` and `porcelain` API.

In this case, maybe an additional simple default parameter `bool updateHEAD=true` would do the trick. Or do you have another idea in mind - maybe by using inheritance or something?

scunz:
> In this case, maybe an additional simple default parameter bool updateHEAD=true would do the trick.

Guess this is the least painfull approach. So I'd say: go ahead :-)

libGitWrap/ObjectCommit.hpp
scunz: "Quickly" actually means a few hours (At least it took a few hours to add them at the tail and adjust MGV to this change).

I'd say: If you have the time to do this right, then please do it. Considering the dependencies on MGV this change will have, it should probably be done once and for good.

However, when you're at it: You could clean it up a little more and remove the TLS_RESULT default parameter totally. Since the transition of mgv is through, this thingie has actually no real benefit (And exactly this was the reason for putting the `Result&` at the tail of the parameter list instead to its head)...

antis81: Done :ok:! Everything is in feature branches `libGitWrap/refacture/ResultAsFirstParam` and `MacGitverModules/refacture/ResultAsFirstParam`. The new functions from this PR can be changed directly here.

antis81:
Ready to merge?

scunz:
> Ready to merge?

Yeah. I don't see anything that really keeps this from being merged...

PR #19, opened 2013-07-03
Improvements to references

antis81:
- [x] list references
- [x] minor improvements


antis81:
Rebased onto merged PR #20.

antis81:
Should we merge here? This will break MGV build a little though caused by renamed methods.

scunz:
breakage can't be avoided... So let's go for it :-)

PR #21, opened 2013-07-26
Fix converting 'unchanged' status

scunz:
"0 & 0" is never true, thus "Unchanged Files" were no longer reported to the WT/Index views.


antis81:
Uhm, oops. Thanks for pointing that one out :smile:.

PR #22, opened 2013-07-30
Feature: invalidate

antis81:
The invalidate method is able to internally replace the underlying lg2 objects or in other words "invalidate" the wrapper object by assigning a 0 pointer to the member. This allows safe synchronization of Git objects relying on the GitPtr system.

For example:
A rename action on a branch invalidates the underlying `git_reference *` and replaces it with a new one, free'ing the memory and doing the right thing from bottom up to the top. Thus every used `Git::Reference` object used in the UI (or anywhere else) points to the same memory and thus gets the same information.

Note:
External changes e.g. from the terminal or whatever can still not be recognized. This needs to be addressed separately.


scunz:
@antis81, I'd really like this change set to be abandoned.

Anything that is only able to access a GitPtr<> should by no means invoke such an action. It cannot access any `git_xxxx` objects anyway (Since it should have used GitWrap to obtain an object and we don't hand out lg2 internals). Contrary, anything inside GitWrap has far deeper access to the internal object model and can do such tragic changes directly. Actually, only the objects themselves should change the private pointer to the lg2-internal object (Git::Reference in the case of the reference). But I still doubt if that would be the desired behaviour.

I believe that you're implicitly drawing the curtain over some hairy edge cases. It is true that there can be 2 different (GitWrap-) objects in memory that point to the same logical lg2 object (a reference or a submodule i.e.). This can only be solved by keeping a hash table in the `Repository` that remembers all created private objects and manages their lifetime. I'm not even saying to reuse them, just thinking about catching the double existence case. I don't think this is what we want.
However, by freeing the old object like in this patch, we're increasing the likeliness of a dangling pointer in the other object. This isn't wrong, but in the way it's done here, introduces hard to debug failures in GitWrap when the actual problem is in user code.

Further, if we'd use this mechanism to clean out the internal data pointer of GitWrap objects, then we must strongly advise any user code to check for validity of each and every GW-Object prior to it's usage. A thing I'd really like to avoid.

So, let's think a second about what we most probably what to have from the user point of view: Anything connected to a `Repository` whats to know about changes (Let's call them 'events'). It most probably doesn't care about where such an event came from; it rather cares about what kind that event is of; like: "Reference created", "Reference renamed" or "reference target changed". Esp. the target changed event can have many different sources, but the event consumer will (again, most probably) treat them all the same.

This train of thoughts leads to the requirement of a m:n relationship between event sources and event consumers. Some hippocratic post modern developers call this an "observer pattern". Luckily Qt abstracts this away from us and provides us with signal and slots.

As for MGV, I'd suggest to incorporate another level into that: the MGV-Core. With the Repository-wrappers implemented in the core, we already got such a layer in between. GitWrap should provide the basic events and implement the _explicit_ sources for them (Like reference was renamed programatically). The core should catch those events and hand them 1:1 on to whoever listens. The core itself however, should keep track of a list of all references and where they point to. On a refresh-request, the core should reread the reference list and create events accordingly.
The result would be: History-List receives a "Target changed for the 'HEAD' ref". It doesn't have to care if that came from an internal git pull, an internal git merge, an internal git checkout or any of the external counterparts issued on the command line. It could just do it's thing...

What do you think about this?

antis81:
First, this looks like a slight misunderstanding concerning the public accessibility. Everything happening with `invalidate` is internal and private to GitWrap. So this cannot by accident be called from the public GitWrap API.

Appart from that, thinking more about it you're probably right. The use cases for that are rare and implementation might not be worth it. Anyways, I'll sort and add the remaining changesets not yet committed and put this branch abandoned.

Let's take a closer look at the synchronization. The point is:
1. As long as we are "inside an MGV instance", every view will be updated according to the information held by the GitPtr, as soon as the view is redrawn. This comes down to simply trigger a "MainWindow->update()" after each `user action`. The underlying object's data (GitPtr to some GitWrap object and thus to the lg2 data object) has already the correct data available. ( This is why we should never (!) copy data (e.g. path, name, ... ) and get it always directly from the "original" object instead. )
2. When the change comes from "outside MGV" (let's assume a terminal), first lg2 needs to know about that change within the repo / file system. I didn't look into that yet, did you? If lg2 gets the notification it only needs to trigger an event, in case an object pointer was added, deleted or replaced.
3. If an lg2 object is just modified (e.g. a member like "path" changed), it is the same like 1. as well. We only need some trigger to update the view. Once again, the cool thing is, that we always access the "original" data, no copy.

This means, we do not really need to observe changes by additional events, because we "already know" they happened and have the correct data at hand.

Am I missing something?

scunz:
> Am I missing something?

Yes, you're missing some things here:

Your 1. is plain wrong. This isn't how mgv works all over the place. You would have to do a lot of changes to get it that way - and it will be way slower. In order for the history view to work, it needs a graph of all commits, ordered in a special way. The edge nodes of that graph change with every ref change. The history view is very optimized. it actually implements a "calculate as you go" mechanism, which populates only minimalistic data to the cache and even defers the node coloring until absolutely required. Esp. the node coloring is a very expensive task. This can be observed when you're opening a _big_ repository (like qt4.git) and directly move the scrollbar to the very end of the history view. In this case the node coloring has to be done for the whole history at once. If we'd let that caching go and really did only rely on accurate data from GitWrap, then we would have to do a RevWalk for each drawn line including the following node coloring. This would delay navigation a lot.
Another case is the divergence view that we want to implement. It also needs to find a merge base, which in turn requires a revision walk. If we didn't cache that result, we'd have to do that calculation for every drawn reference. Actually, we do know that this data only gets invalid when the target of either of both references changed.
The Index and WorkingTree Views are another example. We really can't build our tree during a painting. We need to have it built and calculated through ahead of the painting. Further, we need very special triggers to rebuild it. A change in the target of the special HEAD reference is only one example.

=> We need to watch for those events, en detail.
=> We don't need what triggered the events.

Point 2 is complex. I'd almost say it's utterly nonsense. This is not due to technical reasons and the problems doesn't even exist on _our_ side of the fence. It's more like: LibGit2 devs have (and always will) in any way refused to implement some file system watching triggers to update. This means, the work to do this is totally on our side. Let's do a short analysis of this: There is a system S (the repository) which has two states in time T1 and T2, T2 being later in time than T1. What we want to know is: Any difference in detail between the two states. At the moment T2 is in effect, T1 is no longer available. This means that we can only get the differences if we have detailed information of T1 in memory (=cached). This is by the way also contrary to your 1. point.

=> We already have a semi working code that does all of this via FileSystem Watchers. It's unrelyable code, which is disabled and should be removed.
=> We need a central point that caches the contents of the repository. This is also our chance to use libgit2 the way it was designed to be used: With only very short lived objects.
=> This code should trigger the same events as normal GitWrap operations would do (If a special handling is required at all)

Both of that invalidate your summary-point 3.

The real point is: We do have views that need to act upon changes in very different ways. We're planning to implement more of those.

Hope this makes sense and doesn't disenchant you that much. :-)

antis81:
> => We already have a semi working code that does all of this via FileSystem Watchers. It's unrelyable code, which is disabled and should be removed.

Yeah, did not forget that :+1:. Might be the only way to really watch changes.

> => We need a central point that caches the contents of the repository. This is also our chance to use libgit2 the way it was designed to be used: With only very short lived objects.

This is two in one statements. To the first one I'd say: "Yes, and it is almost there. Just see MGV->RepoManager." The last one I'm not quite convinced, that this is the right way for ALL objects. But at least for commits we could think of storing just the OID's or such. But for simplicity I'm no friend of copying out data into "smaller" objects in the first place. And if so, these again need to be held central -> so it comes out the same way.

> => This code should trigger the same events as normal GitWrap operations would do (If a special handling is required at all)

Hm, do you talk about centralizing `actions`? Well, one step after the other :smile:.

No disenchantment on my side. As long as we search for the "right" solution, everything's cool :smile:.

PR #23, opened 2013-08-04
fixed pointer comparison

antis81:


scunz:
Oh, nice typo :-) However, the change to the destructor is redundant (but doesn't harm either) :+1:

antis81:
Thanks, this actually made some headache, cause the const ref operator compares `d == o.d` :smile:. Wasn't sure about the destructor - good to know.

scunz:
Yeah, anything virtual in a base class is also virtual in it's derviat (As long as the derivat is correctly overloading). i.e. `virtual void foo( int bar )` will make `void foo(int)` in a derivat virtual, but not `void foo(char*)`...

PR #25, opened 2013-08-30
Allow Git::ObjectId to be turned into short strings

scunz:
Sometimes (most of the time) we don't want all the 40 numbers of a full SHA. Instead of creating it fully and then cutting the unwanted part, allow to tell libGitWrap how much we want to.


scunz:
Haven't had a chance to even test this code more than compiling it. But indeed: it also looks good to me too. So let's simply get it in...

PR #26, opened 2013-08-31
Remove obsolete class IndexEntry

antis81:
The underlying `index_entry` struct is private to libgit2 and should not be accessible directly from the public API. For editing, the public methods provided by `Git::Index` should be used. Maybe the API is incomplete and need to be extended for this purpose.


scunz:
> The underlying index_entry struct is private to libgit2

No. It's clearly documented as being public. And we don't use the _private content_ of libgit2.

> and should not be accessible directly from the public API.

Why not? Okay, I'd agree that we should not force users of libGitWrap to use a libgit2 structure. But we need the `Git::IndexEntry` class. We should instead copy all of libgit2's `git_index_entry` data into public (probably!) members of `IndexEntry`

> For editing, the public methods provided by Git::Index should be used.

There are none.

> Maybe the API is incomplete and need to be extended for this purpose.

Yes, actually. A method to "commit" a changed index entry into libgit2 is missing.

scunz:
> > and should not be accessible directly from the public API.
> 
> Why not? Okay, I'd agree that we should not force users of libGitWrap to use a libgit2 structure. But we need the Git::IndexEntry class. We should instead copy all of libgit2's git_index_entry data into public (probably!) members of IndexEntry

I actually saw, that we don't publicly use the `git_index_entry` but still forward-define it. I will remove that declaration in my next cleanup phase (soonish)...

antis81:
> I actually saw, that we don't publicly use the git_index_entry but still forward-define it. I will remove that declaration in my next cleanup phase (soonish)...

Yep, actually we are not using the IndexEntry class anywhere and I still think we don't need it. Everything having to do with that is encapsulated through `Git::Index`. The topic came up with the discussion on how to realize the `stage area`. If really needed in the future we must be aware of, that we are using copied objects which should only live short to avoid inconsistency.

scunz:
I do remember the progress of creating the `IndexEntry` class. Well, we seem indeed to have _currently_ very limited usage for accessing the index's entries. This will, however, change when implementing things like "partial stage". I.e. we need at least to query each entry's in-index SHA (diff index to WD will do that for us, in the first step). In a second step we need to apply a on-the-fly generated diff to the content of the in-index SHA, hash the result and then fill a IndexEntry with it's data and add that SHA to the index.

Most of that is probably true for the `TreeBuilder` and it's `TreeEntry` class.

antis81:
Ah, ok. I forgot about the partial stage topic. So I'll close this one and keep the class with your changes.

scunz:
Okay, let me see, if I can come up with some romantic doxygen poetry instead of this. :-)

PR #27, opened 2013-08-31
Fixing compare operators of Git::Reference

antis81:
The first implementation of object comparison of two `Git::Reference` objects is wrong and error prone (sorry for that). Especially as we are working with smart pointers, we need to worry about when an object is marked equal. This PR adds a simple fix for comparing references of Git::Reference objects (sometimes I hate language :smile:). Question here: Should we compare the `d` member here?

To compare the destination (symbol or OID) of references, the new method `compare` should be used. This wraps exactly `git_reference_cmp` from lg2.

Note: This enforces a correction to the behaviour of the MGV module `RefsViews`, currently relying on the == operator.

### DIFF
--- a/libGitWrap/Reference.cpp
+++ b/libGitWrap/Reference.cpp
@@ -77,6 +77,19 @@ namespace Git
         return !( *this == other );
     }
 
+    /**
+     * @brief   Compares two reference objects.
+     *          Note: References of different types are considered to be different.
+     *
+     * @param   other the reference to compare with
+     *
+     * @return  0 if pointing to the same destination, otherwise an ordered value
+     */
+    int Reference::compare(const Reference &other) const
+    {
+        return git_reference_cmp( d->mRef, other.d->mRef );
+    }
+
     bool Reference::isValid() const
     {
         return d;
--- a/libGitWrap/Reference.cpp
+++ b/libGitWrap/Reference.cpp
@@ -69,10 +69,7 @@ namespace Git
 
     bool Reference::operator==(const Reference &other) const
     {
-        if( d && other.d )
-            return git_reference_cmp( d->mRef, other.d->mRef ) == 0;
-
-        return false;
+        return this == &other;
     }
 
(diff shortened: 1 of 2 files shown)


libGitWrap/Reference.cpp
```diff
@@ -69,10 +69,7 @@
 
     bool Reference::operator==(const Reference &other) const
     {
-        if( d && other.d )
-            return git_reference_cmp( d->mRef, other.d->mRef ) == 0;
-
-        return false;
+        return this == &other;
```
scunz: > Question here: Should we compare the d member here?

Indeed, it should be _just_: `return d != other.d;`. `this` and `&other` can be different, even if their actual data pointers are the same. i.e. your implementation would deem this code to return `false`:

```
bool foo( Reference& r ) {
    Reference r2 = r;
    return r == r2;
}
```

scunz:
Basically I am fine with this change, but I'm not sure at all that I understand the motivation behind this change.

> To compare the destination (symbol or OID) of references, the new method compare should be used. This wraps exactly git_reference_cmp from lg2.

Does this mean, that `git_reference_cmp` compares two different refs (like `refs/heads/a` and `refs/heads/b`) as equal, if both are pointing either to the same symbol or the same SHA!?

antis81:
> Does this mean, that git_reference_cmp compares two different refs (like refs/heads/a and refs/heads/b) as equal, if both are pointing either to the same symbol or the same SHA!?

Yeah, exactly that. The result codes from lg2 have the following meanings:
- -1 -> ref types differ, first ref is symbolic
- 1 -> either ref types differ, second is symbolic ( or the position of first difference in oid or string )
- >=1 -> position of first difference in ref names or oid's
- 0 -> matching:
  - For symbolic refs, the target names match
  - For direct refs, the oid's match

This is used in the RefsViews to show branches pointing to HEAD marked "lighter orange" and the active branch in "orange". This is actually very useful.

scunz:
Then let's use this code :-)

PR #28, opened 2013-08-31
Cleanup QList usage and some stuff

scunz:
@antis81, could you have a short look over this. It does:
- Move all libGitWrap object forward declarations into a central place in GitWrap.hpp
- Changes all remaining usages of QList with QVector
- Does proper typedef declarations for all Lists
- Makes qDebug() macros published code instead of inline
- Do Qt-Includes centraly (the ones we need for API in GitWrap.hpp, most others privately in .cpp)
- Doesn't include <QDebug> in a global scope (early on). Most of our code doesn't use QDebug and where we do, we can as well include it explicitly.

### DIFF
--- a/libGitWrap/ChangeListConsumer.hpp
+++ b/libGitWrap/ChangeListConsumer.hpp
@@ -21,14 +21,12 @@
 
 namespace Git
 {
-    struct ChangeListEntry;
 
     /**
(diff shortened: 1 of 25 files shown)


antis81:
Great changes :+1:. I don't see anything wrong here.

PR #29, opened 2013-08-31
Improve the Index / IndexEntry classes

scunz:
- [x] Enumerator for stages
- [x] Add stage parameter to `getEntry`
- [x] Advance to an almost full wrapper of libgit2
- [x] romantic doxygen poetry

### DIFF
--- a/libGitWrap/CMakeLists.txt
+++ b/libGitWrap/CMakeLists.txt
@@ -124,6 +124,8 @@ SET( SRC_FILES
     IFetchEvents.cpp
     Index.cpp
     IndexEntry.cpp
+    IndexConflict.cpp
+    IndexConflicts.cpp
     Object.cpp
     ObjectBlob.cpp
     ObjectCommit.cpp
@@ -156,6 +158,8 @@ SET( PUB_HDR_FILES
     IFetchEvents.hpp
     Index.hpp
     IndexEntry.hpp
+    IndexConflict.hpp
+    IndexConflicts.hpp
     Object.hpp
     ObjectBlob.hpp
     ObjectCommit.hpp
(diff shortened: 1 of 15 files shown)


scunz:
@antis81, could you have a look at the docs that I added here. I think that it isn't complete yet; but it should cover the things for which we originally decided to spend time on this issue.

The `IndexEntry` and `IndexConflict` classes are still very undocumented. `Index` itself is now fully documented and even has an introduction part in the class documentation.

Actually, I don't feel like writing yet more documentation for stuff for which I still can not determine its usefulness or even check for bugs... So I'd appreciate to get this merged with some kind of note along the line: "It ought to be enough to _read_ everything that an index provides"

antis81:
In conclusion, IMHO the docs are very informative and readable and I don't see anything missing here.

@scunz Why you say the docs are incomplete? At least for the Git::Index it's very "romantic doxygen poetry" :smile:. Big thanks for writing all that.

scunz:
Readability could actually still be improved. But I _currently_ don't mind that - And they're still incomplete because I did not add anything substantial to `IndexEntry` and nothing to the new `IndexConflict` / `IndexConflicts` classes.

Btw.: Did you notice how shamefully `IndexConflicts` and `Index` share the same data class? :-)

scunz:
Oh, forgot: I'll fix the typos when I'm back in London. Thanks for spotting them. Hopefully this will be tomorrow. Actually, I'm trying to find a flight or train that is on _my_ schedule both in terms of time and budget since about 4 hours...

antis81:
Yeah, safe journey :airplane: - May the good lord bless your time over there.

PR #30, opened 2013-09-17
get the name prefix of a git reference

antis81:


scunz:
_Ping_

I think, you wanted this for the References code and forgot to adjust the code to it, right? This is looking good. Just get it in :-)

antis81:
Yeah, right - two things:
1. The code currently doesn't use this -> not breaking anything
2. The reason for that is, that this won't work correctly on remote refs and still "forgets" part of the prefix - but this is another problem to be solved later on :).

scunz:
I might help defining exactly what the "prefix" is. I don't think git itself has that term. But it's still nice to have this... :-)

antis81:
Ok, so I'll delete this branch also to make room for new ideas :).

PR #33, opened 2013-10-05
Add a reference name analyzer

scunz:
Adding an analyzer for reference names.

### DIFF
--- a/Tests/CMakeLists.txt
+++ b/Tests/CMakeLists.txt
@@ -6,6 +6,7 @@ QT_PREPARE(Core -Gui)
 SET(SRC_FILES
 
     TestRepository.cpp
+    TestRefName.cpp
 )
 
 INCLUDE_DIRECTORIES(${libGitWrap_includes})
--- a/Tests/TestRefName.cpp
+++ b/Tests/TestRefName.cpp
@@ -370,3 +370,47 @@ TEST(RefName, RemoteBranch) {
     EXPECT_EQ(0, rn.namespaces().count());
     EXPECT_STREQ("home", qPrintable(rn.remote()));
 }
+
+TEST(RefName, CustomRule) {
+    const char* sz = "refs/pull/550/head";
+    const char* szNeg = "refs/heads/master";
+
+    QRegExp re(QLatin1String("^refs\\/pull\\/([1-9][0-9]*)\\/head$"));
+    int id = Git::RefName::registerExpression((void*)0xAABBCCDD, re);
+
+    Git::RefName rn = Git::RefName(QLatin1String(sz));
+    EXPECT_TRUE (rn.isValid());
+
+    EXPECT_FALSE(rn.isBranch());
+    EXPECT_FALSE(rn.isSpecial());
+    EXPECT_FALSE(rn.isTag());
+    EXPECT_FALSE(rn.isCommitNote());
+    EXPECT_FALSE(rn.isHead());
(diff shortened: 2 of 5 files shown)


scunz:
@antis81, I think this is implemented and covered with suitable tests now. There's also a "tiny" class documentation. Given that docs, do you think we should sub-document all of the methods?

antis81:
@scunz Yes, I also looked over this and everything looks good. Although the `cond ? a, b: false;` syntax gave me some trouble in understanding :smile:. But you just overloaded the ',' operator right?

> Given that docs, do you think we should sub-document all of the methods?

Think this is good as is.

antis81:
Btw.: As currently not used, we can safely remove the `prefix` method and replace it with this approach. I see no reason in keeping it. When it is merged, I can adjust the rename dialog to this.

scunz:
> But you just overloaded the ',' operator right?

Uhm, well no, I'm actually just "using" it. Thus, it's more that `return a ? b, c : d;` is like:

``` c++
if (a) {
    b;
    return c;
}
else {
    return d;
}
```

scunz:
So, I'm getting this in... Wrt. prefix() method, you cannot simply replace it with this code. This code has no way to give you everything that is before the "scope".

antis81:
Thanks for explaining the comma operator. I didn't know that, but assumed exactly what you wrote.

> Wrt. prefix() method, you cannot simply replace it with this code. This code has no way to give you everything that is before the "scope".

What I mean: The `prefix` method is actually dead code at the moment and I'd rather use the `RefNameAnalyzer` instead. In addition, in case of a namespace I am not sure what prefix returns. In case of a remote it returns `refs/remotes`, but we need `refs/remotes/origin`.

scunz:
> I didn't know that, but assumed exactly what you wrote.

Yeah, it is one more totally useless language feature of C++ that I have finally found a viable use case for - after dealing with C++ for at least 20 years ^^

> What I mean: The prefix method is actually dead code at the moment and I'd rather use the RefNameAnalyzer instead. In addition, in case of a namespace I am not sure what prefix returns. In case of a remote it returns refs/remotes, but we need refs/remotes/origin.

Now, that you're saying it _that clear_, I tend to agree with you. I actually had more faith in it; but you're obviously right. So we should get rid of it (again) and add this feature to the `RefName` class.

I'd say: Let's define it with 2 new functions as:
-   `localName()`
  - returns empty string if neither isBranch() nor isTag() returns `true` else
  - returns scopeName() + '/' + name() if isScope() returns `true` else
  - returns name()
  - as a hint: After this, branchName() and tagName() can be simplified to "return empty or localName() if isXXX returns true"; XXX being Tag or Branch.
-    `fullName()`
  - Just for completeness: returns the FQRN as set in the constructor.
-   `scopePrefix()`
  - returns fullName() with the localName() removed.
  - NOTE: Contrary to other getters in `RefName` this includes a trailing slash unless the slash would be the only character.

The only difference in the implementation of scopePrefix() vs. Reference::prefix() would be that we can now better determine what scope() is. Do you want to setup a PR for that change or should I do so?

antis81:
> The only difference in the implementation of scopePrefix() vs. Reference::prefix() would be that we can now better determine what scope() is. Do you want to setup a PR for that change or should I do so?

Please - if you don't mind - feel free to implement these last three methods also. Thanks in advance :heart:.

scunz:
No problem, I will add those then.

PR #34, opened 2013-10-06
Continue RefName analyzer

scunz:
More features as discussed


scunz:
Actually, I will need to add more doxygen, just to silence it's warnings :-)
But that's a job for another day ^^

PR #39, opened 2013-10-27
Refactoring some GitWrap internals

scunz:


scunz:
I'm merging this without any further reviews. I've reviewed the _whole_ change set twice and I've been testing it in my trees for a while. I wish there were lots of tests to confirm nothing broke.

Though, this introduces a lot of style mistakes... I'll fix (and port to lg2 style) incrementally.

PR #40, opened 2013-10-28
Tests: Incrementally build up a Test suite

scunz:
This branch is currently meant to be rebased upon development regularly and to follow development closely.


scunz:
There is no reason to not have these merged right away, is it?

PR #52, opened 2014-01-05
feature: commit changes to repository

antis81:
- [x] implementation of the commit operation -> commits contents of the index to the repository
- [re-planned] implementation to amend a commit
- [x] Spice it up with some tests

### DIFF
--- a/libGitWrap/CMakeLists.txt
+++ b/libGitWrap/CMakeLists.txt
@@ -158,6 +158,7 @@ SET( SRC_FILES
     Operations/BaseOperation.cpp
     Operations/CheckoutOperation.cpp
     Operations/CloneOperation.cpp
+    Operations/CommitOperation.cpp
 
     Operations/Private/WorkerThread.cpp
 )
@@ -205,6 +206,7 @@ SET( PUB_HDR_FILES
     Operations/CloneOperation.hpp
 
     Events/Private/RemoteCallbacks.hpp
+    Operations/CommitOperation.hpp
 )
 
 SET( PRI_HDR_FILES
--- a/libGitWrap/CMakeLists.txt
+++ b/libGitWrap/CMakeLists.txt
@@ -204,9 +204,10 @@ SET( PUB_HDR_FILES
     Operations/BaseOperation.hpp
     Operations/CheckoutOperation.hpp
     Operations/CloneOperation.hpp
+    Operations/CommitOperation.hpp
+    Operations/Providers.hpp
 
     Events/Private/RemoteCallbacks.hpp
-    Operations/CommitOperation.hpp
 )
 
 SET( PRI_HDR_FILES
--- a/libGitWrap/Commit.cpp
+++ b/libGitWrap/Commit.cpp
@@ -121,6 +121,57 @@ namespace Git
         return repo.lookupCommit( result, ObjectId::fromRaw(commitId.id) );
     }
 
+    /**
+     * @brief           Creates a new commit and adds it to the repository.
+     *
+     *                  The HEAD reference and, if it exist, the current branch
+     *                  are updated and point to the new commit.
+     *
+     * @param[in,out]   result  a result object; see @ref GitWrapErrorHandling
+     *
+     * @param repo      the target Git repository
+     *
+     * @param tree      the root tree object within the commit
+     *
+     * @param message   the commit message
+     *
+     * @param author    the original commit author's signature
+     *
+     * @param committer the committer's signature
+     *
+     * @param parents   the commit's parent commit(s)
+     *
+     * @return          the created commit object on success or an invalid object
+     */
+    Commit Commit::create(Result &result, Repository& repo, const Tree &tree, const QString &message,
+                          const Signature &author, const Signature &committer, const ObjectIdList &parents)
+    {
+        if (!result) return Commit();
+        if (!repo.isValid() || !tree.isValid())
+        {
+            result.setInvalidObject();
+            return Commit();
+        }
+
+        Repository::Private* rp = Private::dataOf<Repository>(repo);
+        git_signature* gitAuthor = Internal::signature2git(result, author);
+        git_signature* gitCommitter = Internal::signature2git(result, committer);
+        QString branchName = repo.HEAD(result).name();
+        const git_oid** constParents = Internal::ObjectIdList2git(result, parents);
+
+        if (!result) return Commit();
+
+        git_oid commitId;
+        result = git_commit_create_from_oids( &commitId, rp->mRepo, branchName.toUtf8().constData(),
+                                              gitAuthor, gitCommitter,
+                                              NULL, message.toUtf8().constData(),
+                                              Internal::ObjectId2git_oid(tree.id()),
+                                              parents.count(), constParents );
+        delete[] constParents;
+
+        return repo.lookupCommit( result, ObjectId::fromRaw(commitId.id) );
+    }
+
     Tree Commit::tree( Result& result ) const
     {
         GW_CD_CHECKED(Commit, Tree(), result);
--- a/libGitWrap/Commit.cpp
+++ b/libGitWrap/Commit.cpp
@@ -107,7 +107,7 @@ namespace Git
         git_signature* gitAuthor = Internal::signature2git(result, author);
         git_signature* gitCommitter = Internal::signature2git(result, committer);
         QString branchName = repo.HEAD(result).name();
-        const git_commit** constParents = Internal::CommitPrivate::commitList2git(result, parents);
+        QScopedArrayPointer<const git_commit*> constParents( Internal::CommitPrivate::commitList2git(result, parents) );
 
         if (!result) return Commit();
 
@@ -115,8 +115,7 @@ namespace Git
         result = git_commit_create( &commitId, rp->mRepo, branchName.toUtf8().constData(),
                                     gitAuthor, gitCommitter,
                                     NULL, message.toUtf8().constData(), Private::dataOf<Tree>(tree)->o(),
-                                    parents.count(), constParents );
-        delete[] constParents;
+                                    parents.count(), constParents.data() );
 
         return repo.lookupCommit( result, ObjectId::fromRaw(commitId.id) );
     }
@@ -157,7 +156,7 @@ namespace Git
         git_signature* gitAuthor = Internal::signature2git(result, author);
         git_signature* gitCommitter = Internal::signature2git(result, committer);
         QString branchName = repo.HEAD(result).name();
-        const git_oid** constParents = Internal::ObjectIdList2git(result, parents);
+        QScopedArrayPointer<const git_oid*> constParents( Internal::ObjectIdList2git(result, parents) );
 
         if (!result) return Commit();
 
@@ -166,8 +165,7 @@ namespace Git
                                               gitAuthor, gitCommitter,
                                               NULL, message.toUtf8().constData(),
                                               Internal::ObjectId2git_oid(tree.id()),
-                                              parents.count(), constParents );
-        delete[] constParents;
+                                              parents.count(), constParents.data() );
 
         return repo.lookupCommit( result, ObjectId::fromRaw(commitId.id) );
     }
--- a/libGitWrap/Commit.cpp
+++ b/libGitWrap/Commit.cpp
@@ -55,6 +55,18 @@ namespace Git
             return otCommit;
         }
 
+        const git_commit** CommitPrivate::commitList2git(Result& result, const CommitList &list)
+        {
+            if (!result) return NULL;
+
+            const git_commit** ret = new const git_commit *[list.count()];
+            for ( int i=0; i < list.count(); ++i )
+            {
+                ret[i] = dataOf<Commit>(list.at(i))->o();
+            }
+
+            return ret;
+        }
     }
 
     GW_PRIVATE_IMPL(Commit, Object)
--- a/libGitWrap/Commit.cpp
+++ b/libGitWrap/Commit.cpp
@@ -71,6 +71,56 @@ namespace Git
 
     GW_PRIVATE_IMPL(Commit, Object)
 
+    /**
+     * @brief           Creates a new commit and adds it to the repository.
+     *
+     *                  The HEAD reference and, if it exist, the current branch
+     *                  are updated and point to the new commit.
+     *
+     * @param[in,out]   result  a result object; see @ref GitWrapErrorHandling
+     *
+     * @param repo      the target Git repository
+     *
+     * @param tree      the root tree object within the commit
+     *
+     * @param message   the commit message
+     *
+     * @param author    the original commit author's signature
+     *
+     * @param committer the committer's signature
+     *
+     * @param parents   the commit's parent commit(s)
+     *
+     * @return          the created commit object on success or an invalid object
+     */
+    Commit Commit::create(Result &result, Repository& repo, const Tree &tree, const QString &message,
+                          const Signature &author, const Signature &committer, const CommitList &parents)
+    {
+        if (!result) return Commit();
+        if (!repo.isValid() || !tree.isValid())
+        {
+            result.setInvalidObject();
+            return Commit();
+        }
+
+        Repository::Private* rp = Private::dataOf<Repository>(repo);
+        git_signature* gitAuthor = Internal::signature2git(result, author);
+        git_signature* gitCommitter = Internal::signature2git(result, committer);
+        QString branchName = repo.HEAD(result).name();
+        const git_commit** constParents = Internal::CommitPrivate::commitList2git(result, parents);
+
+        if (!result) return Commit();
+
+        git_oid commitId;
+        result = git_commit_create( &commitId, rp->mRepo, branchName.toUtf8().constData(),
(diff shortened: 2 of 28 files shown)


antis81:
:question: Should we better move the `Index::commitToRepository` method contents to the operation class?

scunz:
Some comments:
1.  The name `commitToRepository` implies that it is called on a bare index and one has to give a repository to it. We need to write both variants (if we do at all, see below): `commit` for the associated repository and `commitToRepository` for bare indices.
2.  Before we start implementing such high level functions that short circuit so many paths,  we should think about what we want (need) to implement. That means, we have to ask some questions, first:
   
   **What is a commit?**
   A commit is an object that connects a hierarchy of trees to a list of zero to n other commits giving them a unique message and an author/timestamp and a committer/timestamp. In git terminology we're calling the combination of name + e-mail + time stamp a _signature_. A commit cannot be transformed into another type of object.
   We can conclude from that: An operation to form a commit needs to have the following parameters: The root of the tree hierarchy, the message, the author signature and the committer signature and a list of zero to n commits to use as parents.
   A user of libGitWrap might need this functionality at this low level, i.e. to implement rebasing, amending, importing and other stuff.
   
   **What is an index?**
   In the context of commit creation, we can reduce an index to: A flat list of named blob objects that are actually contained in a hierarchy. Thus, an index can be created from and converted into a hierarchy of tree objects (Each tree object being a list of blobs and subordinated trees).
   We can conclude from that: There is no direct relation between an index and the creation of a commit. In the first place, we can create methods that convert between trees and indices (Which we already have: `readTree` and `writeTree`). The result of `writeTree` can be used as input to the commit operation.
   
   **Do we have other sources for the tree in a commit operation?**
   Yes, we do: The user might want to use an existing tree (i.e. Create a `foo` branch with all commits from `master` but only containing the path `libFoo/`. This is the subtree functionality in core git). The user might also want to use a `TreeBuilder` to create the tree objects.
   To conclude: In the first place we must implement creating a commit using a tree; however, if we do the same thing like in the checkout operation, with several subclasses, we can make use of our deeper knowledge (namely that libgit2 can itself create a tree out of an index bypassing the need to create temporary libGitWrap objects).
   
   **Does the commit operation need to do anything else?**
   Not necessarily, at least not the basic one. If we go further and implement the operation mechanisms right in all places (Which would be ideal for new code like this), then all methods in the core objects should just return operations. Like `Index::commitOperation()`, which would return an operation suitable to create a commit for this index - or `TreeBuilder::commitOperation()`, which would return an operation suitable to create a commit for the built tree.
   However, here we have several such complex input parameters. See below.
3.  As I did with the `CheckoutOperation`, the `CommitOperation` should be splitted into several ones.
   The `BasicCommitOperationPrivate` would provide a protected method that does the commit based on all information provided above under the bullet _what is a commit_. The `BasicCommitOperationPrivate::run()` method would just call that.
   Then there could be a `CommitIndexOperation`, which would turn the `setTree` method private and provide a `setIndex` instead. It would then overload the `prepare()` method and inside it: turn the given index into a tree.
4.   Since the commit operation has at least two such input parameters, we could as well use functor alike objects to amend the functionality of the commit operation (This is an alternative idea to some of the points in 2. and 3.):
   
    There would be only one `CommitOperation`, but instead of providing methods like `setTree` and `setParents`, it would provide `setTreeProvider()` and `setParentsProvider()`. Both of these methods set a functor-like object that is called during a generic flow. This would be way more flexible than my above suggestion.
    Such a parent provider could look like:

``` C++
class CommitParentsProvider {
public:
    virtual QVector<ObjectId> parents() const = 0;
    virtual bool prepare() { return true; }
    virtual bool finalize(const ObjectId& commitId) { return true; }
};
```

```
`prepare()` would be called before the commit is created. If the parents are user provided, nothing would happen. If the the `CommitOperation` was created from a `Reference` object, prepare() would peel the reference into a commit and set the result as parent. An `AmmendCommitParentsProvider` instead would just copy the parent list from the given commit.
`finalize()` would be called after the commit is created. In case the commit is based on a reference and the reference shall be updated, this method would update the reference. Maybe this should be a functor on its own (Not sure about that, actually).

The same applies to `CommitTreeProvider` objects: A basic version would just return the user-set value. Advanced versions could use `prepare()` to turn an index or tree builder into a tree.
```

antis81:
From this point it is actually possible to create a commit the first time :smile: :rocket:.

Some uncommitted test code:

``` c++
    Git::Repository gitRepo = repo->gitRepo();
//    Git::CommitOperation opCommit;
//    opCommit.setTreeProvider( gitRepo.index(r).writeTree(r) );
//    opCommit.setParentProvider( gitRepo.HEAD(r).peeled(r, Git::otCommit).id() );
//    opCommit.setMessage( ui->textCommitMessage->toPlainText() );

    Git::Index index = gitRepo.index(r);
    Git::Signature sign = Git::Signature::defaultSignature(r, gitRepo);

    if (r)
    {
//        r = opCommit.execute();
        Git::CommitList parents;
        Git::Commit headCommit = gitRepo.HEAD(r).peeled(r, Git::otCommit).asCommit();
        parents << headCommit;
        Git::Commit::create( r, gitRepo, index.writeTree(r), message(), sign, sign, parents );
    }
```

@scunz Need more help with the operation.

:bulb: Btw. shouldn't we define operations in MGV? I Think we should provide only the base classes in GitWrap and create a flexible interface for those.

scunz:
> Btw. shouldn't we define operations in MGV? I Think we should provide only the base classes in GitWrap and create a flexible interface for those.

Actually, no. These operations belong into GitWrap. They're pretty low level and let us avoid writing the same code over and over again :)

scunz:
> From this point it is actually possible to create a commit the first time :smile::rocket:.

btw, RB is just about to change how this works in libgit2 :)

scunz:
Nice additions. But as a hint: To be more fool prove, I tried to make all the setters in the operations fail, if the operation is already started.

antis81:
The "commit operation" is working. Any ideas how to get the pointer creation of `IndexCommitProvider` more clean?

scunz:
Actually, I don't really get why half of this is done that way it is done :angel: So, a few questions about the design:
-   What is the purpose of having only one provider to the operation? If there was only one way to do it, then we could just move the setters back into the operation, right?
  
  I was under the impression, that the purpose would be to have at least the 2 variable parameters to the commit operation separated out of it. Thus it would have to be two providers: One for the list of parents (which could be implemented as "read a ref, commit on top of that and update optionally the ref", "produce a merge of foo and bar" or even "give birth to a new brach") and another one for the tree that will actually _become_ the commit's content. This data could come from a tree builder object, an index object or could even be simply an already existing tree.
-   I'm not sure that it's right to make that providers private.
  
  If someone comes up with something that we cannot even remotely think about, why shouldn't they be allowed to inherit the providers and do their crazy thing themselves?
-   Why is there a method to set defaults? Shouldn't the defaults be set, uhm, _by default_? :)

antis81:
> Why is there a method to set defaults? Shouldn't the defaults be set, uhm, by default? :)

Good question :) - Done.

> What is the purpose of having only one provider to the operation? ...

We have to setup a tree and parents for each commit operation, right? Parents must come from the target repository, the tree can come from outer space :space_invader:. The meaning stays the same, does it? We could provide a list of CommitProviders, but in the first place we're creating a commit on top of other commits (or none for the initial commit). This works for merge scenarios as well as normal commits. Sure I can think of some "commit to new branch" functionality. But wouldn't that just increase complexity of the Workflow?

> I'm not sure that it's right to make that providers private.

Well, that's phylosophical .... I'd rather keep it private unless it is really needed to be part of the public API. That keeps things more easy for now.

scunz:
> That keeps things more easy for now.

I learned to hate this expression. Because most of the time it just translates to "Uhm, I know that there's something incompatible awaiting; but I just don't want to care".

> We have to setup a tree and parents for each commit operation, right?

Yes, and they are not related to each other in any way.

> Parents must come from the target repository, the tree can come from outer space :space_invader:.

Almost right. For sure, parents must come from the target repository. But if the tree is a real tree and not something that will be turned into a tree, it (and all of its blobs and subtrees) has to be part of the parent repository, too.

So. The idea behind the operation and its providers is: lego, actually. A 1-to-1 transformation from UML to code, if you want.

The task is, to create a new commit. To solve this task, we implement a `CommitOperation`. This is an _actor_ in a UML sequence diagram. But we soon realise that this task consists of at least 3 tasks: 1. Preparing the tree(s) to be committed 2. Telling what parents to associate. 3. Creating the actual commit object in the ODB.

Now, instead of making the CommitOperation over complex, we take 1. and 2. and turn them into separate actors in the sequence diagram. This way, the CommitOperation can simply call out to the other two actors at the right point in the flow.

In fact, these new actors just have to follow a given pattern, so each type of them becomes exchangeable. We can write a dozen different ways to prepare the trees and we can write a few different ways to figure out the parent list. And most important: We write a dozen + a few classes instead of writing a dozen X a few classes.

That's also one reason, I'd like these classes to be public. Just imagine the side effects of:

``` C++
class TreeBuilder{
// ...
public:
    operator TreeProviderPtr() const;
};

class Index {
// ...
public:
    operator TreeProviderPtr() const;
};

class CommitOperation {
public:
     CommitOperation(TreeProviderPtr, ...);
//...
};
```

In user-land code, there is suddenly no difference between:

``` C++
Index i; CommitOperation co(i);
// and:
TreeBuilder b; CommitOperation co(b);
```

Though, we don't have to write anything special for that in the CommitOperation.

antis81:
@scunz Thanks for the detailed explaination. I wasn't thinking of reusing the code. So I splitted everything up in the public API again. It's almost finished, but some changes are still needed ...

Two questions I'm still not convenient with:
- ...Provider::repository() -> Should I add a setter/getter like for the simple info?
- Index...Provider::Index -> This currently prohibits use of the operators -> Should I put it in the ctor instead of a setter?

scunz:
> Index...Provider::Index -> This currently prohibits use of the operators -> Should I put it in the ctor instead of a setter?

I'm not sure whether I understand this :)

antis81:
> I'm not sure whether I understand this :)

Sorry, was happy that it works in the first place and tired of writing ... :). This relates to these lines libGitWrap/Index.cpp:L87 and above. This could be cut down to two lines code by using the operator TreeProviderPtr() in the ctor of the operation. But we need to call setIndex(). But ... with your comment [above](https://github.com/macgitver/libGitWrap/commit/7f801766ceb9142e2a3ebdd3460bd53290c234f4#libgitwrap-index-cpp-P106), this problem may become obsolete. I'll check that later in the evening.

antis81:
@scunz Please forget the question. The following commits will fix everything.

antis81:
Ok ... this goes more in the right direction, eh? :smile:

Next I plan to make `Git::Repository` itself a the ParentProvider. Thus I'd suggest something like this:

``` c++
class ParentProvider
{
public:
ObjectIdList provideParents() = 0;
ObjectId provideHeadParent() = 0;
};

class Git::Repository
public:
  class ParentProvider
  {
    Repository repository() { return *this; }
    ObjectId provideHeadParent() { return HEAD().resolveToObjectId(); }
    ObjectIdList provideParents() { 
    /* future will tell:
    for merging, we surely need some reference names to provide the tip commit id's ... */
    }
  };

public:
  operator ParentProviderPtr() { return ParentProviderPtr( new Repository::ParentProvider() ) }
};
```

Now the creation of the CommitOperation for committing the staged contents itself burns down to:

``` c++
Git::Repository repo = /* get the repo */
CommitOperation op( repo, repo.index(r), "my commit message" );
op.execute();
```

Agree'd? The Index won't be a ParentProvider anymore and only provides the `Git::Tree` (operator will be removed). I think this is what we really want. Your comment with the repository brought me to that solution -> thanks a lot for helping out with that :+1:.

antis81:
Yep - thinking the same. But how to differ the parents in the producer (namely the Repository now)? In the first place I'm glad I finally understood what you wanted - it's great :smile:

scunz:
> Now the creation of the CommitOperation for committing the staged contents itself burns down to:
> 
> ``` C++
> Git::Repository repo = /* get the repo */
> CommitOperation op( repo, repo.index(r), "my commit message" );
> op.execute();
> ```
> 
> Agree'd? The Index won't be a ParentProvider anymore and only provides the Git::Tree (operator will be removed). I think this is what we really want. Your comment with the repository brought me to that solution -> thanks a lot for helping out with that :+1:.

If you're going that route _that far_, then it makes sense to go it to the end, too. `HEAD()` already returns a `Reference` object. Currently, this can be a `BranchRef`, a `TagRef` or a `NoteRef`.

Why not create a `ParentProvider` class that accepts (in 3 constructors): `Reference`, `BranchRef` and `TagRef`? For the `Reference` case (private constructor), it would refuse to accept it, if it is not a Branch or Note. Then add a static method `fromRef(Reference&)` that uses the private constructor (To be able to return failure).

This would make it possible to commit on top of everything that should allow to commit (branches and notes) except raw OIDs. The latter should be possible too, since it's perfectly fine to commit to a detached head :)

So, if we add a `headBranch()` method to `Repository`, that returns HEAD as `BranchRef` or fails, then committing to HEAD is only slightly different:

``` C++
CommitOperation op( repo.headBranch(r), repo.index(), "Foo the Bar!");
```

But, actually it should be rather:

``` C++
CommitOperation::Ptr op = repo.headBranch(r).commitOp(repo.index(), "Foo the Bar!");
```

`commitOp` could be a virtual method in `Reference` as well (which simply fails for TagRefs). Then it would even be:

``` C++
repo.HEAD().commitOp(repo.index(), "Foo the f**king Bar now")->execute();
```

scunz:
> Yep - thinking the same. But how to differ the parents in the producer (namely the Repository now)?

Simple: Don't. If the Repository _keeps_ being a parentProvider, it should as well have a commitOp() method. To be used like:

``` C++
repo.commitOp("Just, Foo the Bar, NOW!")->execute();

// i.e.: 
// CommitOperation::Ptr Repo::commitOp(QString msg)
// { return HEAD().commitOp(index(), msg)); }
```

(Which would simply take the only sane defaults: Parent=HEAD; Tree=Index)

> In the first place I'm glad I finally understood what you wanted - it's great :smile:

You're welcome :)

scunz:
So much `fooing the foocking bar`... I should run for president in foobarkistan :)

antis81:
Well, everything should be in the right place now - after so much foo'ing and bar'ing and moving code around :tada: :smile:.

:question: The CommitOperation::Ptr aproach is not working out of box, cause we inherit `BaseOperation` from `QObject`. Should I make it inherit from `QSharedData` instead?
:question: Repository::commitOperation() and the reference equivalent does not return gracefully in every situation. Should we `Result` as an additional parameter?
:question: On execution of a `Operation` we create a second `Result` object. Shouldn't we pass it through instead? Or is it not possible because of threading issues?

scunz:
> Well, everything should be in the right place now - after so much foo'ing and bar'ing and moving code around :tada::smile:.

Hey, after all, this was a very constructive discussion and a nice incremental development process. With lots of good inputs from all sides and keeping rethinking the thing until it eventually became near to perfect... I _really **really** wish_ it was like that everywhere... Esp. at some customer places :)

As for your questions:

> The CommitOperation::Ptr aproach is not working out of box ...

Sad, but true. But, no, actually `QSharedData` is not an option. Remember that these things are meant to be run in a background thread _optionally_. Anyway, if all goes wrong, this could be done with a non-intrusive shared pointer like `QSharedPointer` - though I don't like it one bit.
So, a QExplicitlySharedDataPointer, QSharedDataPointer or QSharedPointer, give us no bargain. We cannot use `operator->` on these, unless we create a static instance that does nothing... Not a good idea, IMHO.
So, actually, there's two options:
a) Accept the fact that operation-creators can fail _and return `NULL` on failure_. After all the trouble we're going through everywhere to ensure that we don't use such API, this somehow reminds me of the :snake: and the :apple:...
b) Write a wrappers that act as Value-Semantic-Classes like we do everywhere else. This is more tricky here, since we'd be dealing with a `QObject` behind the scenes. Actually, nothing that I'd really like to do.

So. I'm not sure. Maybe it's best to keep it as it is...

For the other questions:
The result object inside the operation has to be used in the worker thread. Actually, the `execute` method should hand out that object.

antis81:
Hm ... assumed leaving everything as is right now. Are we ready to merge?

scunz:
Except for missing tests, I don't see a reason to _not_ merge this :)

antis81:
Ok, I'm going to write some tests down this weekend and merge afterwards. The providers for the `TreeBuilder` are also still missing. Should I add them as well?

scunz:
I think that we should add them, _if_ we add sufficient tests for them :)

antis81:
Added the stuff to TreeBuilder -> no tests yet. Let me add them before merging.

antis81:
I want to merge this. Together with the history features this PR adds a lot value to MGV. I'll write more tests by the time, but I really wish to get V0.1 features finally shipped :boat:. Would be nice, if we could some help getting an installer package ready.

scunz:
Installers once used to work. Not sure about current state. But this probably involves getting all the tiny bits in their proper places again for all operating systems. I mean things like "relative dylib references on macosx", RPath on Linux and so on and so on...

PR #53, opened 2014-03-09
removed deprecated call of repository()

antis81:
Removed the deprecated call to repository(). Seems to be :ok: so far.


scunz:
Yeah, we (at least I) often deprecate some stuff with newer one, then actually port everything and just don't go the extra mile to clean up accumulated mess. Thanks for spotting this. I don't dare to merge it, since I cannot test anything. But feel free to do so, if MGV compilation goes through :)

PR #54, opened 2014-03-15
CheckoutOperation

antis81:
IIRC the checkout operation was implemented as a draft with most stuff implemented, but untested. So this PR is meant to complete it, test it and maybe re-discuss some parts.

:question: Is it right adding the export macros? Or was it left out on purpose?


scunz:
> Is it right adding the export macros? Or was it left out on purpose?

I don't think that they're required, but it cannot harm much either :)

antis81:
I added them because the compiler doesn't know `CheckoutTreeOperation` and such. So .... leaving it like that.

scunz:
If at all, the linker should bark about that. If it's the compiler, then something else is wrong. I.e. you're trying to use the CheckoutTreeOperation without the header included...

However, I'm still thinking that these are not required (at least technically): When do you really need to tinker with the objects these operations are bound to operate on?

What I mean in more detail. The workflow (based on a Tree here) should be:
- One gets a Tree object from somewhere
- Ask the Tree to create a CheckoutOperation (It creates a CheckoutTreeOperation and returns it as CheckoutBaseOperation)
- Eventually _run_ the operation
- Get the result from it (a task of the base class, actually)

The additional methods that a CheckoutTreeOperation provides (compared to the CheckoutBaseOp), is a getter() for the Tree (which we already know, since THIS tree created the op in the first place).

antis81:
Ahhh, crap. I was just about to start it over like with the `CommitOperation` discussion.

There isn't a checkout operation on a commit yet. So I would "copy" `Reference::CheckoutOperation` in the first place to make it "do something".

When thinking this further ... We can use the same mechanism we have now with the operation providers, can we? Thus I could either merge the `CommitOperation` - which I don't want to do, as for missing tests. Better I'd like to close this very PR and add the stuff to PR #52.

Meaning:
- Make `Git::Commit`a TreeProvider and give it the operators.
- Realize the `CheckoutOperation` in the flavor of `CommitOperation`, but - like you stated above - return the `CheckoutBaseOperation`
- Replace `Commit::checkout()`with `Commit::checkoutOperation()`

antis81:
PR closed as it doesn't make much sense. Further improvements for the checkout operation go to branch `nf/f/commit-changes`

Edit: Ooops - corrected nonsense overflow - :dizzy: :smile:

scunz:
:)

PR #55, opened 2014-10-08
upgraded to libgit2 v0.21.1

antis81:
@scunz Could you please check especially the change to using `git_buf` in pathes. Otherwise everything should be fine now :smile:.

Again for the `git_buf` I'd like to use a QScopedPointer. But it would need to implement the custom destructor. Ok?

### DIFF
--- a/libGitWrap/Commit.cpp
+++ b/libGitWrap/Commit.cpp
@@ -161,11 +161,11 @@ namespace Git
         if (!result) return Commit();
 
         git_oid commitId;
-        result = git_commit_create_from_oids( &commitId, rp->mRepo, branchName.toUtf8().constData(),
-                                              gitAuthor, gitCommitter,
-                                              NULL, message.toUtf8().constData(),
-                                              Internal::ObjectId2git_oid(tree.id()),
-                                              parents.count(), constParents.data() );
+        result = git_commit_create_from_ids( &commitId, rp->mRepo, branchName.toUtf8().constData(),
+                                             gitAuthor, gitCommitter,
+                                             NULL, message.toUtf8().constData(),
+                                             Internal::ObjectId2git_oid(tree.id()),
+                                             parents.count(), constParents.data() );
 
         return repo.lookupCommit( result, ObjectId::fromRaw(commitId.id) );
     }
@@ -390,12 +390,23 @@ namespace Git
         }
     }
 
+    /**
+     * @brief           Create a branch on this commit.
+     *
+     * @param[in,out]   result  A result object; see @ref GitWrapErrorHandling
+     *
+     * @param[in]       name    the branches reference name
+     *
+     * @param[in]       force   if true, creation is forced (an existing branch will be moved)
+     *
+     * @return          the created reference
+     */
     Reference Commit::createBranch(Result& result, const QString& name, bool force) const
     {
         GW_CD_CHECKED(Commit, Reference(), result);
 
         git_reference* ref = NULL;
-        result = git_branch_create(&ref, d->repo()->mRepo, name.toUtf8().constData(), d->o(), force);
+        result = git_branch_create( &ref, d->repo()->mRepo, name.toUtf8().constData(), d->o(), force, NULL, NULL );
 
         if (!result) {
             return Reference();
--- a/libGitWrap/Config.cpp
+++ b/libGitWrap/Config.cpp
@@ -44,68 +44,76 @@ namespace Git
     QString Config::globalFilePath()
     {
         QString filePath;
-        char path[ 2048 ];
+        git_buf path = {0};
 
-        int rc = git_config_find_system( path, 2048 - 1 );
+        int rc = git_config_find_system( &path );
         if( rc >= 0 )
         {
-            filePath = QString::fromLocal8Bit( path );
+            filePath = QString::fromLocal8Bit( path.ptr );
         }
 
+        git_buf_free( &path );
         return filePath;
     }
 
     QString Config::userFilePath()
     {
         QString filePath;
-        char path[ 2048 ];
+        git_buf path = {0};
 
-        int rc = git_config_find_global( path, 2048 - 1 );
+        int rc = git_config_find_global( &path );
         if( rc >= 0 )
         {
-            filePath = QString::fromLocal8Bit( path );
+            filePath = QString::fromLocal8Bit( path.ptr );
         }
 
+        git_buf_free( &path );
         return filePath;
     }
 
     Config Config::global()
     {
-        char path[ 2048 ];
+        git_buf path = {0};
 
-        int rc = git_config_find_system( path, 2048 - 1 );
+        int rc = git_config_find_system( &path );
         if( rc < 0 )
         {
+            git_buf_free( &path );
             return Config();
         }
 
         git_config* cfg = NULL;
-        rc = git_config_open_ondisk( &cfg, path );
+        rc = git_config_open_ondisk( &cfg, path.ptr );
         if( rc < 0 )
         {
+            git_buf_free( &path );
             return Config();
         }
 
+        git_buf_free( &path );
         return PrivatePtr(new Private(cfg));
     }
 
     Config Config::user()
     {
-        char path[ 2048 ];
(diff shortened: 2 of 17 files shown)


scunz:
Overall, this looks straight forward. I didn't check for _all_ cases that the signature parameter is actually allowed to be `NULL`. Did you?

However, there's 3 things that I don't really like:
- Changing the code format without having a formal specification how the code should look. Esp. removing braces around one line conditions, _is_ a bad idea. They're there for a reason: Maintainability and better readability.
- Adjusting to a newer library version and not exposing the new functionality inside changed methods is not a good idea. If the objective is to just make it somehow compile again, then that's okay. But esp. when we're talking about a library like gitWrap, new functionality should be exposed right from the start. Otherwise it will be forgotten...
- The `git_buf` handling as it is done is error prone and ugly (for C++ :smile:). Using a `QScopedPointer`, even with a custom deleter is just as ugly. If it is correctly wrapped inside a C++ class, there's no need for any ugly code outside of that class...

antis81:
> Overall, this looks straight forward. I didn't check for all cases that the signature parameter is actually allowed to be NULL. Did you?

Yep, I did. It tries to create a "default signature" for the log file. If that cannot be generated, it errors out by setting the return code. So everything good with this one.

> Changing the code format without having a formal specification how the code should look. Esp. removing braces around one line conditions, is a bad idea. They're there for a reason: Maintainability and better readability.

Ok, I think I can discipline myself in not doing such things in the future and become a "black belt clean coder" :8ball:.

> Adjusting to a newer library version and not exposing the new functionality inside changed methods is not a good idea. If the objective is to just make it somehow compile again, then that's okay. But esp. when we're talking about a library like gitWrap, new functionality should be exposed right from the start. Otherwise it will be forgotten...

For this exact reason I made up this PR and I'm going to keep this practice in the future. But actually figuring all the new features out, wrapping them and finally implement into MGV (or even other apps) is a very time consuming task. Especially when taking a look at our very limited resources. So yeah, the first goal is to make it compileable again and then focus on the update stuff we urgently need in MGV. Not to forget, still we gain from some bugfixes. I really wish we'd get more people involved in the project to better being able to achieve these goals. But anyways, we could make up a short list of what we really need for MGV to be implemented for the next version. Ah well, I'll do it in redmine and add a link in the comments below. This won't work always though ...

Btw.: What really sucks about libgit2 is, that people there often do such _really stupid_ decisions like renaming essential structures (in this version e.g. "oid" to "id" or "git_checkout_opts" to "git_checkout_options"). Just why?! If you look into the changes here, it is the reason why most of the stuff didn't compile anymore. This is really unnecessary :zap: :angry:. Sorry, needed to let that beast out :smile:.

> The git_buf handling as it is done is error prone and ugly (for C++ :smile:). Using a QScopedPointer, even with a custom deleter is just as ugly. If it is correctly wrapped inside a C++ class, there's no need for any ugly code outside of that class...

Yeah, that's why I wanted you to take a look at it - thanks :smile: :exclamation:. So wrapping `git_buffer` into a `Git::Buffer` object it should be. I think we should combine it with`QBuffer` then. But better I make up some code, so we can discuss that here. Just wait until this evening. I really need to get a :haircut: and cleanup the house before going on (it took me 2h to bath the keyboard and get the lemonade out of it - like :new:) :smile:.

scunz:
> Btw.: What really sucks about libgit2 is, that people there often do such really stupid decisions like renaming essential structures (in this version e.g. "oid" to "id" or "git_checkout_opts" to "git_checkout_options"). Just why?! If you look into the changes here, it is the reason why most of the stuff didn't compile anymore. This is really unnecessary :zap::angry:. Sorry, needed to let that beast out :smile:.

I'm not trying to defend anyone here, but you got to remind a few things:
- First of all, `libgit2` is an officially unreleased software package under development. I say officially unreleased, but it should actually mean _there is no final release yet_. Take it as an pre-alpha stage software. There is no stability promise on the API, of which they occasionally take advantage.
- The `oid` to `id` change is indeed arguable.
- And last, but not least, remember the fact that _API-Design_ is a different art than _development_. And this one really needs an _artist_ to get it somehow right from the start. Luckily, it's just `C` code, with `C++` it is usually much worse (i.e. I created a generic wrapper for Lua, it took me like 2 hours in C to get  a decent API for it. I had to the the same in C++ and made the mistake to not just take the C code... I tried to do it the C++-way. Took me more than 2 days - and the API is much less optimal than the C one).

That being said, I'd really advise to start following the `master` branch again. It's much less work in the end.

scunz:
> Ok, I think I can discipline myself in not doing such things in the future and become a "black belt clean coder" :8ball:.

The code is a total mess anyway, IMHO. I should have closely followed the libgit2 style right from the start (I do that for all of my projects now, and it's way easier to switch).

scunz:
So, we're coming to the `git_buf` now... What does it have in common with `QBuffer`? I can't see a relation.

What we need is a very basic C++ style API around it:

``` C++
class StringBuf {
private:
    git_buffer buf;
public:
    String() { memset(&buf, 0, sizeof(buf)); }
    ~String() { git_buffer_free(&buf); }
    operator git_buffer*() { return &buf; }
    operator QString() { return /*...*/; }
private:
    StringBuf(const StringBuf&);
    StringBuf operator=(const StringBuf&);
};
```

This allows us finally, to take some output from libgit2 and input it into another function without the non-loss-less QString conversation.

antis81:
Thanks for providing almost complete implementation. Implemented it in GitWrapPrivate (because of `git_...` objects are not being exported).

Also not replaced the `QString` operator through `const char*`. When implementing both, the operator becomes ambiguous. Another variant would be implementing a `constData()` method. But I think that one is ok, is it?

Also replaced the missing `Results` and deprecated everything else. Thus, it stays fully compatible and yet it's a bit of a less mess :+1:.

scunz:
What about a `Buffer::toString()` then? Maybe, not inlined, even. I actually want to hide the call to `fromUtf8` from the compiler, so it has to make a call into the Buffer class...

antis81:
Not sure if I understand you right. First, in this context `fromLocal8Bit` is used. Don't know if this is correct, but looks valid. And isn't it already hidden like you want it? Or do you want being able to write `QString("...");` instead of `QString::fromUtf8("some string");` again?

Ahem, in Qt4 it was possible to set the encoding globally using `QTextCodec::setCodecForCStrings()`, but it was removed in Qt5 API - for good reason.

scunz:
> Not sure if I understand you right. First, in this context fromLocal8Bit is used. Don't know if this is correct, but looks valid.

It isn't. Even if this once was good enough, git repositories are generally said to be completely UTF-8 encoded (except for commit messages, where an encoding can be explicitly specified).

> And isn't it already hidden like you want it? Or do you want being able to write QString("..."); instead of QString::fromUtf8("some string"); again?

I want to write `buf.toString()` and have the compiler not inline all the gory details, hoping that a) the binary size will not explode and b) that at least the assembly sequence to trigger the recoding of the string is centralised and the CPU holds it in cache when it's heavily used. One might call that paranoia, but I've suffered so much from badly designed software that I tend to keep an eye open for small trivial changes that make a huge difference in sum.

> Ahem, in Qt4 it was possible to set the encoding globally using QTextCodec::setCodecForCStrings(), but it was removed in Qt5 API - for good reason.

Yeah, gone for good. However, this feature was for a different purpose. It was to be used when strings have been _implicitly_ converted to UCS-2 (i.e. QString(const char*) etc.)

antis81:
> I want to write buf.toString() and have the compiler not inline all the gory details ...

Hm, we should implement a `setTextCodec()` to the buffer then as well. Default is "utf8" of course. `toString()` does the conversion then by using this codec. Simple as that.

EDIT:
Please note, that the current implementation uses this functions for path names. Such are not related to the ODB directly! This is why I think `fromLocal8Bit` is in good use here.

antis81:
I'll merge this one manually and "fast-forward" it. It's pretty straight forward and we already have a branch following the new `libgit2/master`. This one will build with `lg2 v0.21.1` - but not on master. I made these changes on my local machine though - following later in our `development` branch.

PR #56, opened 2014-10-12
String wrapper to convert between QString and lg2

antis81:
Introducing `Git::Internal::String` wrapper class to simplify conversion from and to QString.
- Should it be exported to the public interface? If so, the GitWrap API should change to take this class instead of a QString.

### DIFF
--- a/libGitWrap/GitWrap.cpp
+++ b/libGitWrap/GitWrap.cpp
@@ -40,7 +40,7 @@ namespace Git
 
             for( unsigned int i = 0; i < arry->count; i++ )
             {
-                sl << Internal::StringHelper( arry->strings[ i ] );
+                sl << GW_StringToQt( arry->strings[ i ] );
             }
 
             git_strarray_free( arry );
@@ -216,7 +216,7 @@ namespace Git
 
             QString path;
             if (df->path) {
-                path = Internal::StringHelper::convert(df->path);
+                path = GW_StringToQt(df->path);
             }
             return FileInfo(path, ObjectId::fromRaw(df->id.id), df->size, FileModes(df->mode),
                             false, (df->flags & GIT_DIFF_FLAG_VALID_ID) != 0);
--- a/libGitWrap/GitWrap.cpp
+++ b/libGitWrap/GitWrap.cpp
@@ -98,7 +98,7 @@ namespace Git
 
         //-- String ------------------------------------------------------------------------------>8
 
-        String::String(const QString& str, QTextCodec* codec)
+        String::String(QString str, QTextCodec* codec)
             : mStr(str)
             , mCodec(codec)
         {
@@ -115,9 +115,28 @@ namespace Git
             mStr = convert( str, size, codec );
         }
 
-        String::operator QString()
+        QString String::convert(const char* str, int size, QTextCodec* codec )
         {
-            return mStr;
+            return codec ? codec->toUnicode( str, size )
+                         : QString::fromUtf8( str, size );
+        }
+
+        QString String::convert(const char* str, QTextCodec* codec )
+        {
+            return codec ? codec->toUnicode( str )
+                         : QString::fromUtf8( str );
+        }
+
+        const char* String::convert()
+        {
+            mConvertedStr = toArray();
+            return mConvertedStr.constData();
+        }
+
+        QByteArray String::toArray() const
+        {
+            return mCodec ? mCodec->fromUnicode( mStr )
+                          : mStr.toUtf8();
         }
 
         String::String(const String& other)
@@ -131,34 +150,20 @@ namespace Git
             return *this;
         }
 
-        QString String::convert(const char* str, int size, QTextCodec* codec )
+        String::operator const char*()
         {
-            return codec ? codec->toUnicode( str, size )
-                         : QString::fromUtf8( str, size );
-        }
-
-        QString String::convert(const char* str, QTextCodec* codec )
-        {
-            return codec ? codec->toUnicode( str )
-                         : QString::fromUtf8( str );
-        }
-
-        const char* String::convert(const QString& str, QTextCodec* codec )
-        {
-            return codec ? codec->fromUnicode( str ).constData()
-                         : str.toUtf8().constData();
-        }
-
-        String::operator const char*() const
-        {
-            return convert( mStr, mCodec );
+            return convert();
         }
 
         String::operator char*()
         {
-            return const_cast<char*>( convert( mStr, mCodec ) );
+            return const_cast<char*>( convert() );
         }
 
+        String::operator QString()
+        {
+            return mStr;
+        }
 
         //-- StrArrayWrapper -------------------------------------------------------------------- >8
 
--- a/libGitWrap/GitWrap.cpp
+++ b/libGitWrap/GitWrap.cpp
@@ -80,7 +80,7 @@ namespace Git
          */
         QString Buffer::toString() const
         {
-            return StringHelper( buf.ptr, buf.size );
+            return GW_StringToQt( buf.ptr, buf.size );
         }
 
 
--- a/libGitWrap/GitWrap.cpp
+++ b/libGitWrap/GitWrap.cpp
@@ -80,8 +80,7 @@ namespace Git
          */
         QString Buffer::toString() const
         {
-            return mCodec ? mCodec->toUnicode( buf.ptr, buf.size )
-                          : QString::fromUtf8( buf.ptr, buf.size );
+            return String( buf.ptr, buf.size );
         }
 
         Buffer::Buffer(const Buffer& other)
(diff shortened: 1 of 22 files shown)


libGitWrap/GitWrap.cpp
```diff
@@ -97,6 +96,70 @@ namespace Git
...
+                         : QString::fromUtf8( str, size );
+        }
+
+        QString String::convert(const char* str, QTextCodec* codec )
+        {
+            return codec ? codec->toUnicode( str )
+                         : QString::fromUtf8( str );
+        }
+
+        const char* String::convert(const QString& str, QTextCodec* codec )
```
scunz: All branches in this method return pointer to invalid memory!

antis81: Ooops, thanks! Hm, should we store a `QByteArray` then instead of a `QString` and do the conversion in the c'tor? Or even store the `const char*`?

antis81:
@scunz Fixed the conversion bug, that accessed invalid memory. This one allso affects the formerly introduced `Git::Buffer`.

Some more thoughts:
1. When using the wrapper throughout GitWrap, it seems to having a positive performance impact on MGV. I'll commit that one soon. It is mixed with heavy changes to the `CloneOperation`, through following the master - I'll leave that one out for the moment.
2. I think this one is suitable very well to use in the public API. We should make use of it whenever passing a QString to GitWrap.
3. While thinking about passing the `QTextCodec`, I think we don't really need that at all and always convert to utf8. Speaking more generally, when the codec changes, it changes at exactly one place.
4. From the point before, I think it makes really sense to either name it `Git::Utf8String` or better `GitWrap::StringHelper` or something like that.

Please let me know your thoughts. Is it complete bullshit to build upon such a class?

libGitWrap/GitWrap.cpp
```diff
@@ -97,6 +96,75 @@ namespace Git
...
+        }
+
+        String::String(const String& other)
+        {
+            *this = other;
+        }
+
+        String& String::operator =(const String& other)
+        {
+            *this = other;
```
scunz: This is an endless recursion, isn't it?

libGitWrap/GitWrap.cpp
```diff
@@ -97,6 +96,75 @@ namespace Git
...
+        }
+
+        String::operator const char*()
+        {
+            return convert();
+        }
+
+        String::operator char*()
+        {
+            return const_cast<char*>( convert() );
```
scunz: Returning a non-const pointer to const-data is dangerous.

However, since you're actually doing no caching of any intermediate results (esp. not reusing them). Guess what happens in the following code:

``` C++
void test(const char* a, const char* b);
void test2(const String& s) {
    test(s, s);
}
```

Given the current code, `a` will be a pointer to invalid memory.

libGitWrap/GitWrap.cpp
antis81: I think so. The only purpose of this is to privatize the copy consstructor. Could you tell me on how to do it right, please?

libGitWrap/GitWrap.cpp
antis81: Well, good point. Though this case should be rare, it is likely to happen. I was thinking about something like this

```
const char* String::convert()
{
  if ( mConvertedStr.isEmpty() )
  {
    mConvertedStr = toArray();
  }

  return mConvertedStr.constData();
}
```

Would that be ok? The conversion only is performed once (could event be in the c'tor then). The problem is the `operator char*`, which I didn't want to have originally - it is used currently only in one place in GitWrap (think it was the DiffList class??). Thus, it would become a _constant String object_. Maybe this is the best way doing it.

scunz:
> While thinking about passing the QTextCodec, I think we don't really need that at all and always convert to utf8. Speaking more generally, when the codec changes, it changes at exactly one place.

That's what I was trying to say right from the start. The codec **is** fixed to UTF8 - and there's only exactly one place where the whole git concept doesn't honour this (the commit message, which is allowed to have an arbitrary encoding). And even there it is discouraged and only exists to support the ancient, legacy behaviour of south eastern developers. (Git devs would express it a bit more polite, but that's basically it).

> From the point before, I think it makes really sense to either name it Git::Utf8String or better GitWrap::StringHelper or something like that.

Ideed, but see below.

> When using the wrapper throughout GitWrap, it seems to having a positive performance impact on MGV.

**I challenge that!** First of all, we're talking about _ns_ here, which would be barely noticeable from within MacGitver. At most the tests could be effected. However, looking at the code. Even if:
- _all_ of it was inline, _and_
- the _codec_ thing would have been completely removed, _and_
- the compiler actually doesn't go beyond its threshold (which is definitely ZERO in debug builds and IIRC for GCC 50 in release builds) and hence manages to inline everything _and_
- the running _CPU_ has a very small cache pipeline _and_
- the _data cache_ problem stays as it is _and_
- foreseeable threading issues with this code are _not_ protected against.

even _in that optimal case_ (where almost all of your code would be removed by the compiler), you'd still be running exactly the same assembly code.
But since none of that is true, there's more code to run (Taking more time and causing more cache misses on the executable memory pages), many more stack frames to produce (resulting an increased amount of memory writes and thus cache flushes).

There's however, one case when a class _like this_ might improve speed. That's if you actually do _not_ use Qt to do the conversation, but use a local UTF8-Conversation method (which is easy to do). In that case, you might be able to avoid one cache miss in a system that is both under very high memory pressure (or has a very high thread count) and has a very long code trail that uses GitWrap exclusively. Not a very realistic scenario, anyway.

> I think this one is suitable very well to use in the public API. We should make use of it whenever passing a QString to GitWrap.

If you remove the codec stuff, this class is essentially a wrapper around `QString::toUtf8()` and `QString::fromUtf8()` - which have actually nothing to do with libgit2. So, unless you're planning to remove Qt support from libgit2, it doesn't make much sense to wrap Qt in our public API.

Overall, you might be better off with:

``` C
#define GITWRAP_CHARSTAR_TO_QT(x) QString::fromUtf8(x)
#define GITWRAP_QT_TO_CHARSTAR(x) (x).toUtf8().constData()
```

libGitWrap/GitWrap.cpp
scunz: Generally spoken, I'd express that in the actual API (class definition), basically by making every non-modifing method `const` and the `mConvertedStr` member `mutable`.

non-modifing is meant in the semantic case, not literally. It _has_ to change `mConvertedStr` occasionally, but that's why it is mutable. This is by the way the same reason, I usually declare the base class for reference counting like:

``` C++
class Base {
public: void addRef() const { mRefCount++; }
public: void delRef() const { if (!--mRefCount) delete this; }
private: mutable int mCounter;
};
```

libGitWrap/GitWrap.cpp
scunz:
``` C++
// until C++03
class X {
private: X(const X&); /* Without providing an implementation */
private: X& operator(const X&); /* Without providing an implementation */
};

// Since C++11:
class Y {
Y(const Y&) = delete;
Y& operator=(const Y&) = delete;
};
```

The `X` class will raise a compiler error, if it is ever copied or assigned _outside_ of `X` and a linker error if used from inside (`void X:test() { X x = *this; }`), unless you're covered by stupid coding rules that force you to implement everything. In that case you'd loose the linker error for internal usage (But you could replace it with a runtime error by throwing an uncatchable exception).

The `Y` class will yield compiler errors for internal and external usage of copy, assignment, new-placement-with-copy, move and move-assign operations (and doesn't allow an implementation of these methods in the first place).

scunz: Oh, and given the literal definitions as I wrote them in the previous example, some people also do:

``` C++
class Foo : private X // or Y for that matter
{
    // cannot be copied, assigned, moved, etc...
};
```

antis81:
@scunz Can you please check the latest changes (especially the last 3 Commits).

First, I removed StringHelper in favor to the macro solution, which I also think now is way better than anything else.

By going through the changes, I stumbled on another assignment of an invalid address in our good old `StrArray` wrappers, caused by `QString::toUtf8().data()`. So I fixed this, by copying the converted QByteArrays into the internal data member.

 The problem is yet, when assigning something like this:

``` c++
void some_method() {
  git_strarray b; // this is a class member in our case
...
  {
    b = StrArray a( validQStringList );
  }
  // at this point 'b' is totally f***ed up, when it shouldn't be - sorry :(
}
```

In our case, b is a "hidden member" - and that's very bad. I think the best way to fix this, is to build a Wrapper around `git_checkout_options`, containing a `StrArray` member. Currently, our `CheckoutOptions` is just an enum and the rest is completely missing.

Question: Will the `git_strarray_free` produce a crash?

Same applies to the `CloneOperation`

antis81:
I'd like to merge this and go on with the update to the `master` and further implementing the `CheckoutOptions` and `CloneOptions` wrappers. Think they will both do good and serve us the ability to finally implement both operations right.

scunz:
@antis81, sorry to be a bit late in this game. Have been a bit sick recently.

However, I don't get what the most recent commits are about at all...

antis81:
@scunz Sorry to hear that. Hope you get well soon.

The changes are relatively simple. I realized your suggestion:

``` c++
#define GITWRAP_CHARSTAR_TO_QT(x) QString::fromUtf8(x)
#define GITWRAP_QT_TO_CHARSTAR(x) (x).toUtf8().constData()
```

But renamed the macros to:
`GW_StringFromQt` and `GW_StringToQt`

Further, there's a 3rd macro `GW_QStringToQByteArray`, producing the utf8-encoded QByteArray. By the way everything works together, now there's really only this one place where to change the encoding - cool thing.

Last not least I did rename `StrArrayWrapper` to simply `StrArray` and in turn the former `StrArray` to `StrArrayRef` - because that's what they really are.

While replacing the macros I noticed, that `CheckoutOperation` and `CloneOperation` are still pretty unfinished stuff. With the latest commits I just prepared everything to finish these two operations.

Finally, I introduced new wrappers for `git_clone_options` (which was just added for lg2 >v0.21) and `git_checkout_options` written down locally and even changed StrArray one more time to better match API semantics. Thus, I also implemented the copy-constructor of StrArray and such make it work right. The stuff is working almost complete, but crashes at one point. Damn, we really need some unit tests to verify the basic structures are always working as intended... Ah well, I simply commit the stuff and we discuss it in a PR. We really need to.

_EDIT:_
Sorry made a mistake. The `GW_QStringToQByteArray` macro is actually called `GW_EncodeQString`.`

PR #57, opened 2014-10-15
follow lg2 master

antis81:
I set this PR up mainly because of an major change in the inner workings of `Git::CloneOperation`. It now uses the available `git_clone_options`.

:checkered_flag: TODO's:
- [x] After merging the `Git::String` wrapper (whatever its final name will be). The changes must be taken here as well.
- [x] Implement wrappers for `git_checkout_options` and `git_clone_options`

### DIFF
--- a/libGitWrap/Base.cpp
+++ b/libGitWrap/Base.cpp
@@ -110,8 +110,8 @@ namespace Git
     }
 
     Base::Base(const Base& other)
-        : mData(other.mData)
     {
+        *this = other;
     }
 
     Base::~Base()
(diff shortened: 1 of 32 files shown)


antis81:
Note: The commits from this PR were rebased onto development to make development easier :innocent:.

antis81:
@scunz Ok, found and fixed the crash caused by our `StrArray` wrappers. We cannot safely just reference the QStringList data. So I decided to not cache the data from the "original" QStringList, but convert it in both directions. Thus, the char array under the hood can even change and is correctly converted back to a QStringList. I'll write some more tests around this though.

Also, I introduced the `CHECK_GIT_RESULT` macro in our test suite. With this, it is for example  easy to see what is wrong with the Commit::create() test (The tree is an invalid object and thus the function fails with exactly this message). For this test we'd have to make up an artificial tree, that gets commited as "initial commit".

With this code, the Checkout- and CloneOperation works again Although both operations need more cleanup to work properly.

PR #58, opened 2014-10-31
forward operation result in finished signal

antis81:
Thinking about, that this doesn't really hurt ... This allows us to review the result of an operation and give the user a helpful hint on what went wrong and what to do. The error messages are very close to the Git commandline.

The alternate is to keep a member of an (asynchronus) operation and request the result within the `finished()` slot. But we can do that still either. This is just an easy way :sun_with_face:


antis81:
Not needed, because we can simply `static_cast` the slot's `sender()`, which is much more effective. I added this now to MGV without having to change anything in the API.

:exclamation: Closing for good.

PR #59, opened 2014-11-14
installers

antis81:
PR to finish the installers for libGitWrap.
- [ ] Create a code gatherer script, so the package creation can be automatically run on the build server.
- [ ] Create a PPA on Launchpad ( https://launchpad.net/~nilsfenner/+archive/ubuntu/macgitver-dev )
  To add the PPA write this in a terminal `sudo apt-add-repository ppa:nilsfenner/macgitver-dev`
- [x] Generate Debian source package for uploading to the PPA.
- [x] Create DEB install package
- [ ] Create an RPM installer for Suse/Redhat distro's.
- [ ] Create a Homebrew package for OSX.
- [x] ~~(NOT NEEDED) Finish the OSX DragNDrop Installer~~
- [x] ~~(USE WiX INSTEAD) Finish the NSIS based installer for Windows~~
- [x] ~~(CANCELLED, BECAUSE NOBODY NEEDS IT)  Finish the WiX based installer for Windows (create an MSM merge-module)~~

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,7 +81,7 @@ ELSEIF( LINUX )
     SET( CPACK_GENERATOR                TGZ TBZ2 DEB RPM )
     SET( CPACK_SOURCE_GENERATOR         TGZ TBZ2 )
 ELSEIF( WIN32 )
-    SET( CPACK_BINARY_NSIS              ON )
+    SET( CPACK_BINARY_WIX               ON )
 ENDIF()
 
 # == Basic information ==
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,7 @@
 
 PROJECT(GitWrap)
+set( PROJECT_MAINTAINER "Nils Fenner" )
+set( PROJECT_CONTACT "Nils Fenner (My personal key) <nilsfenner@web.de>" )
 CMAKE_MINIMUM_REQUIRED(VERSION 2.8.12)
 
 FIND_PACKAGE(RAD-Tools REQUIRED)
@@ -42,3 +44,75 @@ RAD_ADD_DOXYGEN_TARGET(
     ${MGV_GITWRAP_SOURCE_DIR}
     EXCLUDE ${MGV_GITWRAP_SOURCE_DIR}/libgit2
 )
+
+
+#== Packaging ==
+
+# To Create a package go into the build directory
+# and run one of the following commands:
+#
+# - On Mac OSX run: cpack -G "DragNDrop"
+# - On Windows run: cpack -G "NSIS"
+# - On Linux, you have the following options:
+#   1. To create a DEB package, run "make package"
+#   2. To upload the PPA for Ubuntu, enable the "PreparePPA" flag
+#      and (after building) run "make dput" to upload the PPA.
+#      Make sure, the "devscripts" package installed.
+#   3. (NOT IMPLEMENTED YET) -- To create an RPM package run: cpack -G "RPM"
+#
+# This will create the package from the "CpackConfig.cmake" file.
+#
+# You may want to use other package generators like e.g. RPM.
+# To do so, just change the attribute after the "-G" parameter
+
+
+# == Environment ==
+if(NOT CPACK_SYSTEM_NAME)
+    set(CPACK_SYSTEM_NAME "${CMAKE_SYSTEM_PROCESSOR}")
+    if(CPACK_SYSTEM_NAME STREQUAL "x86_64")
+        set(CPACK_SYSTEM_NAME "amd64")
+    endif()
+endif()
+
+# == CPack generstor ==
+IF( APPLE )
+    SET( CPACK_BINARY_DRAGNDROP         ON )
+ELSEIF( LINUX )
+    SET( CPACK_GENERATOR                TGZ TBZ2 DEB RPM )
+    SET( CPACK_SOURCE_GENERATOR         TGZ TBZ2 )
+ELSEIF( WIN32 )
+    SET( CPACK_BINARY_NSIS              ON )
+ENDIF()
+
+# == Basic information ==
+SET( CPACK_PACKAGE_NAME                  "libGitWrap" )
+SET( CPACK_PACKAGE_VERSION               "0.1.0" )
+SET( CPACK_PACAGE_VENDOR                 "http://macgitver.info" )
+SET( CPACK_PACKAGE_CONTACT               "${PROJECT_CONTACT}" )
+SET( CPACK_PACKAGE_DESCRIPTION_SUMMARY   "Qt API for libgit2." )
+SET( CPACK_PACKAGE_DESCRIPTION           "The GitWrap library wraps the libgit2 library for Qt applications." )
+#SET( CPACK_PACKAGE_DESCRIPTION_FILE      "${CMAKE_CURRENT_SOURCE_DIR}/DESCRIPTION" )
+
+
+# == DEB package config ==
+SET( CPACK_DEBIAN_PACKAGE_MAINTAINER        "${PROJECT_MAINTAINER}" )
+SET( CPACK_DEBIAN_PACKAGE_ARCHITECTURE      "${CPACK_SYSTEM_NAME}" )
+SET( CPACK_DEBIAN_PACKAGE_SECTION           "devel" )
+SET( CPACK_DEBIAN_PACKAGE_PRIORITY          "optional" )
+#SET( CPACK_DEBIAN_PACKAGE_DEPENDS          "libgit2 (>=0.21.1)" )
+set( CPACK_DEBIAN_BUILD_DEPENDS             debhelper cmake rad-tools )
+
+# == RPM package config ==
+#SET( CPACK_RPM_PACKAGE_REQUIRES "" )
+## == Source package config ==
+
+
+include( CPack )
+
+if( ${PreparePPA} )
+    # TODO: This should be asked from the distro
+    set( UBUNTU_NAME    "trusty" )
+    set( DPUT_HOST      "ppa:nilsfenner/macgitver-dev" )
+
+    include( cmake/UploadPPA.cmake )
+endif()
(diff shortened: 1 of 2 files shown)


scunz:
@antis81,
1. If you make a proper RAD-Tool out of that cmake script (that can be loaded via the usual `RAD` commands, ... (But what I think you want is actually a place to share that script with the rest of the MGV code, right? Then maybe a Build-Scripts Repository might be better; it could still contain RAD-Tools)
2. What is the purpose of a Drag and Drop installer for Mac OS X ("for a library like libGitWrap")? It could only be a `.framework`-bundle. And this one would have to contain all the Qt framework bundles. Then do the same thing for libHeaven? DnD 10 frameworks for libHeaven; contain the Qt Frameworks in each of them? Think this might not be the best idea to proceed.
   
   Packages for Linux are IMHO okay; Recipes for Mac OS X based package managers might be okay as well. But that's about that: It's okay to feed package managers with instructions (be it binary or via source code). They got support for dependencies and can do "the right thing" most of the time.
3. What exactly is wrong with the windows build? Can you open an Issue or mail me about the actual problem?

scunz:
And the same probably applies to Windows: It doesn't make much sense to produce an installer for Windows, unless it is a MSI-Style "Merge-Package" (`*.msm`) that can be included as a component into a larger MSI installer package...

antis81:
> (But what I think you want is actually a place to share that script with the rest of the MGV code, right? Then maybe a Build-Scripts Repository might be better; it could still contain RAD-Tools)

Hm, ok. A new central scripts repo is a good idea. This has to go to Launchpad as well then as a build dependency.

> What is the purpose of a Drag and Drop installer for Mac OS X ("for a library like libGitWrap")?

Right, there's no purpose here. This should go to the MGV main repo. There's a chance though, that we want to install the GitWrap and Heaven as standalone libraries. Maybe we would be better off with  Homebrew ( http://brew.sh ) here. This works similar to APT.

> And the same probably applies to Windows.

Uhm, right. But MSM (the merge modules mechanism) and NSIS don't like each other. I think there's a better way to get it right. It is mostly, because NSIS makes the deployment more easy by probably just adding another build target, instead of including MSI. Or am I wrong?

> What exactly is wrong with the windows build? Can you open an Issue or mail me about the actual problem?

Basically libgit2 doesn't build like we integrate it. There's a problem including the `http_server.h` header and probably more, but this is what I get. I should mention though, I'm building with `msys/mingw` (on Windows XP, but this shouldn't matter at all).

scunz:
> Hm, ok. A new central scripts repo is a good idea. This has to go to Launchpad as well then as a build dependency.

That's plain stupid then :)
Ain't you allowed to just reference it as a submodule and be done with it?

> There's a chance though, that we want to install the GitWrap and Heaven as standalone libraries. Maybe we would be better off with Homebrew ( http://brew.sh ) here. This works similar to APT.

In that case, yes - this was actually, what I was trying to say. However, I indeed think that anyone building an application for Mac OS X _will_ know exactly, if frameworks or flat libraries are needed. For such a use case, it might be best to provide a simple and easy to run build experience - Since I guess any decent developer will probably do it by themselves.

> But MSM (the merge modules mechanism) and NSIS don't like each other. I think there's a better way to get it right. It is mostly, because NSIS makes the deployment more easy by probably just adding another build target, instead of including MSI. Or am I wrong?

Once you got your head wrapped around the MSI Technology (and the WiX tools), it's quite easy and straight forward. And so much more powerful than all the *nix alternatives... :)

Really, NSIS is more pain in the longer run. The CMake support isn't very good either (IMHO).

> Basically libgit2 doesn't build like we integrate it.

Surprise. Surprise. Who would have guessed? The way I did it, was a quirk, a kludge, a real nasty way to get rid of the ignorance to my improvements. But that's past. It should be possible to do it better nowadays. Better means: Build upon the `CMakeLists.txt` they provide and use their shared library...

> There's a problem including the http_server.h header and probably more, but this is what I get. I should mention though, I'm building with msys/mingw (on Windows XP, but this shouldn't matter at all).

Oh my dear... You're building for that ancient monster? And even using a so much diversified tool chain? Seriously, [nobody uses Wixp anymore](https://escapefromxp.azurewebsites.net/)!
Anyway: The mingw part is more of a problem in my opinion. There're so many different flavours and builds of it out there. Most of them are incompatible with each other. See the struggle Qt-Project.org always has with that stuff. In the end, I very much doubt the usefulness of such a build, unless it is solely based upon the stuff provided by Qt-Project.org. But they actually do not officially support WinXP either. In fact, the new QWebEngine module demands MS VC 12 (aka 2013)...

antis81:
> Build upon the CMakeLists.txt they provide and use their shared library... (libgit2)

Let's leave the past behind then and switch the submodule to the original lg2 repo. That's the best we can do and it will ease the building/testing process.

> Oh my dear... You're building for that ancient monster? ... (WinXP)

Funny game! I'm not using Windows at all and be happy :). It runs in a VM only to build for this OS! The question is: Does it work with that toolchain on Win7/8? If so, MGV should completely build (by copying the library binaries and headers manually to the MGV build dir). Would be great, if we could provide a windows binary as well. If I read right, we can provide a WiX (oh, what a name :laughing:) installer from CMake, right? Using that on windows is the better alternative, yes.

scunz:
> Let's leave the past behind then and switch the submodule to the original lg2 repo.

The repository-copy is just a safety-net. Keep it, you should :)

> I'm not using Windows at all and be happy :). The question is: Does it work with that toolchain on Win7/8?

Well... Yes and no. Windows really is the path to the dark side... Once you're there, you're doomed to stick with it. Should read like: If you provide it once, you'll probably have to _support_ it - and you know how people are: You give them a helping hand and they'll come back tomorrow and grab your whole arm...

scunz:
Oh, and yes: you're right a) WiX is indeed an interesting name for a "program that adds something to another person's intimidate data locations" and b) seems like CMake supports this :)

antis81:
> Ain't you allowed to just reference it (central cmake modules) as a submodule and be done with it?

Yeah, that's actually a good alternative :).

Btw.: What I wrote about the RAD-Tools dependency is completely crap. Sorry for being awkward :innocent:. I'm just learning to deal with Launchpad. When providing the "apt-get install" mechanism, we need to improve our deployment strategy. Because the sources are actually built on Launchpad. Maybe I should make up an issue on its own for that.

Just to remember: Providing binary DEB packages is the more "easy way" and similar to a windows installer. Packages just have to be installed by the user in the right order (libraries first). Problem here: MGV itself doesn't start. 

Do you have some Linux VM? Then please try for yourself:

``` sh
# first run the cmake scripts (I use Qt-Creator)
cd /path/to/RAD-Tools-build && sudo make install
cd /path/to/libGitWrap-build && sudo make install
cd /path/to/libHeaven-build && sudo make install
cd /path/to/MacGitver-build && sudo make install

# run the programm
MacGitver
```

It will fail to start, like I said on the phone. Seems like `libHeaven` breaks it.

antis81:
Oh, and yes: you're right a) WiX is indeed an interesting name for a "program that adds something to another person's intimidate data locations" and b) seems like CMake supports this :)

Ok, then let's go this way. I'll replace NSIS with WiX in the TODO list.

scunz:
@antis81, take a step backwards, a deep breath and then think about this again.

What you really need to do is:
- [ ] Make sure the sources are able to build everywhere you want to deploy. This doesn't correlate to the origin of the sources at all; i.e. just ensure that a full set of sources builds.
- [ ] You need a script to export the sources; preferable to both a tarball and a CR/LF line ending crippled .zip file.
- [ ] Next step is to provide a hosting space for these tarballs.
- [ ] Refer to these tarballs as the "release" source inside your packages and on the web sites..

scunz:
Back in the ol' days, when our Git Lab was still alive, we actually had a jenkins job, which actually did exactly that. Grab the sources from all relevant submodules and daily update a git repository with a _complete_ set of all MacGitver sources in one place.

scunz:
[see here](https://github.com/macgitver/MacGitver/blob/development/scripts/gather-all-code.sh)

antis81:
Thanks for that hint :+1:!

Hm, this is what `UploadPPA.cmake` is doing (well, almost), just before the tarball is generated. But instead of cloning, it copies the local source directory.

I'd prefer to re-write the code gatherer as a cmake module. This way it becomes more platform independent. And ... I have to admit I'm not that good in shell scripting :relaxed:. What do you think?

scunz:
If I get this right, you want to wrap the shell commands from the _gatherer_ inside `ADD_COMMAND` tags in CMake **becuase** you're not good in shell scripting? Does that make sense? Actually, unless you're going to write a complete git wrapper for CMake, there's no other way than running git CLI step-by-step (which is indeed what the gatherer already does).

I'd rather like to have it in a format that can be easily integrated into an automated system and be run automatically...

antis81:
Not exactly. I'd not wrap it, but write it completely as a CMake module (function or a macro). This way it can be run on win / lin / mac the same way, without the need to write separate scripts for each OS ... and without having to cross-compile. Something like this

``` cmake
find_program( GIT git )
if( NOT ${GIT} )
    MESSAGE(STATUS "Git was not found on this system.")
    return()
endif()

function( git_clone_source REPOSITORIES BASE_DIR)
    # we could do a check for existing directories and fetch recursive
    # I'd prefer to delete BASE_DIR instead
    foreach( REPO IN REPOSITORIES )
        execute_process( ${GIT} clone --recursive ${REPO} ${BASE_DIR}/ )
    endforeach()
endfunction()

# ... and so on ...
```

scunz:
This is still wrapping the commands. Except that it will not become cross platform, since it would still require a msys-style bash on windows to deal correctly with the forward slashes... :)

antis81:
Well, when having installed Git on Windows, you have an msys bash anyways :). But the real question is. Why should that be bad? We have to run cmake anyways. So why not using it?

scunz:
Counter question: If you want this kind of bullet-proof, you'll have to write a _huge_ amount of CMake scripts, which is by far not easy. The gain is that a system that is actually not at all meant to be run by any user and is hence just required by the maintainer to execute (or even better be automated in the background), becomes usable on a system which neither you nor me make use of. So, could the time spent on that be spend with a higher valuable output?

scunz:
btw: If you're using Visual Studio (which probably is still like 90-95% of Windows developers), you'd be advised wisely to avoid all kinds of msys in the first place :)

antis81:
Slept over that one and it's hard but true. Supporting a system, that none of the developers uses, does not make much sense. Thus, unless someone wants to support a WiX installer, GitWrap for Windows won't get one.

Using a shell script is a reasonable way to gather the sources then. Though, I can't tell the difference between cmake and shell script in this context. Except shell script is more flexible when replacing the build system in a far away future :).

PR #63, opened 2015-01-05
complete checkout operations

antis81:
Realization of the basic `CheckoutOperations` concept.

Up to now, libgit2 only provides very low-level support for doing checkouts. Thus, we're mostly on our own here and need to design on how this is realized in our API. 

:recycle: I really like the basic idea of defining `Operations`. To get a good restart, I cleaned up and wiped out some unfinished concepts (i.e. `Reference::checkoutOperation()`).

Some very basic scnenarios work (again) - yay :+1:
- checkout HEAD tree
- checkout any commit from the history
- checkout references (also branches)
- checkout the index

TODO:
- [x] Finish API design.
  - [x] Use new `CheckoutCallbacks` API
  - [x] Implement and use `Operation Providers`
  - [x] Implement a virtual `postCheckout` method
- [x] Update the HEAD pointing to a reference, when needed.
- [ ] Add some API doumentation.
- [x] Some basic tests for repo index are added.
- [x] fixed double free in `StrArray`
- [x] Creation of lightweight and normal tags.
- [x] Added (failing) tests for creation of tags. This need work on its own -> handled in new PR.

:exclamation: :memo: **Hint:** This **_WILL_** break GitWrap API. Thus, some MGV features will need adjustment after merged.


antis81:
:+1: Everything just got better. With flowers and butterflies :bouquet: :bouquet: - and brand new bugs :beetle: :grin:. Hard work though :sweat_smile:.

@scunz I know you don't have much time, but ... please ... I am excited to hear your thoughts on this one.

libGitWrap/Events/Private/GitEventCallbacks.cpp
```diff
@@ -213,8 +212,9 @@ namespace Git
         void CheckoutCallbacks::checkoutProgress(const char* path, size_t completed_steps, size_t total_steps, void* payload)
         {
             ICheckoutEvents* events = static_cast< ICheckoutEvents* >( payload );
+            Q_ASSERT( events );
 
-            qreal progress = 100 / total_steps * completed_steps;
+            qreal progress = total_steps ? (100 / total_steps * completed_steps) : 100;
```
scunz: Shouldn't it be 0% progress if no steps are defined?

antis81: The only sane reason I can think of for reporting with `total_steps=0` is, "when nothing needs to be done". Assuming, this is not a lg2 bug, 100% is correct.

scunz: Think that's okay then.

scunz:
Just as a heads up: I saw that CMN is trying to remove the `signature` parameters everywhere. The reasoning is sane, but that will be a major API change in lg2 (see their 2808).

libGitWrap/Events/Private/GitEventCallbacks.cpp
antis81: Thx! Got some time for chatting?

scunz: Yepp :)

antis81: Ok nice, FB then?

(sorry ... I'll delete the last comments)

antis81:
> Just as a heads up: I saw that CMN is trying to remove the signature parameters everywhere. The reasoning is sane, but that will be a major API change in lg2 (see their 2808).

Thanks for the hint. For GW it's not a big deal though. I'll deprecate our calls to keep compatibility as soon as the PR got merged over there. We use "NULL" for both parameters anyways.

libGitWrap/Events/Private/GitEventCallbacks.cpp
antis81: Btw: libgit2 itself stumbled upon this just a few days ago. They set 0 in their example, but I still think 100 is the correct value. I'll simply ask them and maybe change this in a new PR, as I'm about to merge and what's there pretty much looks ok to me. See here: https://github.com/ethomson/libgit2/commit/9af3c4169c51b967579042e392937acbc75da79d

antis81: Ah forget about it! We don't need that: See eadcaa0d1caa99a7b5576cde52676e28093b05d4

PR #64, opened 2015-01-17
Result API Convention

antis81:
Cleanup our API convention.
- [x] In general all methods have to check on the `Result` and directly return, if false.
  - There are very few exceptions on that rule. For example, requesting the `Object.objectId()`, it will return an invalid objectId and set the result object.
- [x] Corrected API to take `Result` as first parameter. (Important: deprecate previous calls!)
- [x] Declare deprecated methods `inline` where possible.
- [x] Document deprecated functions.

:zap: It was not possible to inline the `Object::as...` methods, so I deleted them completely. They must not be used anyways.

### DIFF
--- a/libGitWrap/Repository.cpp
+++ b/libGitWrap/Repository.cpp
@@ -294,7 +294,6 @@ namespace Git
      */
     Index Repository::index( Result& result )
     {
-        GW_CHECK_RESULT( result, Index() );
         GW_D_CHECKED(Repository, Index(), result);
 
         if( isBare() ) {
@@ -330,7 +329,6 @@ namespace Git
      */
     Git::StatusFlags Repository::status(Result &result, const QString &fileName) const
     {
-        GW_CHECK_RESULT( result, FileInvalidStatus );
         GW_CD_CHECKED(Repository, FileInvalidStatus, result);
 
         unsigned int status = GIT_STATUS_CURRENT;
@@ -342,7 +340,6 @@ namespace Git
 
     StatusHash Repository::status(Result &result) const
     {
-        GW_CHECK_RESULT( result, StatusHash() );
         GW_CD_CHECKED(Repository, StatusHash(), result);
 
         git_status_options opt = GIT_STATUS_OPTIONS_INIT;
@@ -369,7 +366,6 @@ namespace Git
      */
     QStringList Repository::allReferenceNames( Result& result )
     {
-        GW_CHECK_RESULT( result, QStringList() );
         GW_D_CHECKED(Repository, QStringList(), result);
 
         git_strarray arr;
@@ -381,7 +377,6 @@ namespace Git
 
     ReferenceList Repository::allReferences(Result &result)
     {
-        GW_CHECK_RESULT( result, ReferenceList() );
         GW_D_CHECKED(Repository, ReferenceList(), result);
 
         Internal::cb_append_reference_data data = { d };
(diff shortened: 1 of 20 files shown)


libGitWrap/Repository.cpp
```diff
@@ -1053,33 +1045,79 @@ namespace Git
...
-            return NoteRef();
-        }
+        // dwim doesn't work for notes
+        return reference(result, QLatin1Literal("refs/notes/") % noteName).asNote();
+    }
+
 
-        return ref.asNote();
+
+    // -- DEPRECATED FUNCTIONS BEGIN -->8
```
scunz: These should probably be inline wrappers in the header, shouldn't they?

scunz:
Hey @antis81, I think you've missed something important here. Could you please have a very deep look at these things:
1. The static `BasePrivate::isValid` method
2. The non static `BasePrivate::isValidObject` method.
3. Follow the flow through the `GW_xxxx` checker macros.

I just wanted to check if this code is still as I designed it, so I've lookup these methods just now. I reckon that it might be hard to spot the architecture in the first place - So I'd definitely suggest to write some documentation on these macros.

However, the design idea was to have both as few macro invocations and as few run-time checks as possible.
The basic idea is, that every non-static method be it const or non-const will need access to the `d`-pointer - and we need to cast it and check it. Thus, each method prolog reads like something along `GW_CD_CHECKED(Commit, DiffList(), result)`. This code's primary objective is to give access to a validated d pointer (That's the `D` and `CHECKED` parts in the name). In this case a `const` version of it (the `C` in the name).

If you follow that chain deeply, i'll notice that internally a `d` pointer is qualified as valid, if:
- It is non-zero.
- If it's virtual validation method determines that it is valid.
- The default implementation for that virtual method is to return `true` / `false` depending on the validness of the given result object. IOW: If the result holds an error, the method simply assumes that the currently tested object is invalid. A nice short cut, esp. since it has no side effects.

If in contrast, the d-pointer is `NULL`, the result will be set to an invalid object. However, this might be a mistake. Actually, `Result::setInvalidObject` should probably refuse to do that if it is invalid itself!

Hence, the `GW_D...` macro should be the first call in every such method and there is no need to check the `Result` object again.

antis81:
@scunz Thanks for writing some lines on this.

> Hey @antis81, I think you've missed something important here. Could you please have a very deep look at these things:
> 1. The static BasePrivate::isValid method
> 2. The non static BasePrivate::isValidObject method.
> 3. Follow the flow through the GW_xxxx checker macros.

Thought a lot about this now and you're right. The only thing wrong with this is the `setInvalidObject`, which could overwrite a previous result. This is unlikely to happen, but not impossible. Basically we only want to provide the "root of all evil" result. Thus, the best way to prevent ever losing this information is to check, if a result already has a value different from 0. If so, the result should not be re-assigned.

This applies for the two `Result::setError` methods and of course `Result::setInvalidObject`. If we should run ever in a situation, where the result needs to be re-assigned, we need to previously call `Result::clear`.

By following our idiom "break on first error", this should practically never happen. So adding the check this is just a nice additional "safey belt".

I'll add the checks to `Result` and remove the additional `GW_CHECK_RESULT`.

antis81:
Looks good to me and I'd like to merge soon.

@scunz Can you check this change ac75e8a4eb8ff4156fa6772a99abd0209d470606 please? The `BasePrivate::isValid` already does a `setInvalidObject` in case it is invalid. In case the result was invalid before, it cannot be re-assigned by accident (see commit 7eddae069b9481a90ea30138694a1bda278d6e3d).

The `Result` is in general now only written once on first error. Setting a result multiple times is ignored. If this is needed, a `Result.clear()` must be called before assigning the new result.

PR #65, opened 2015-02-17
extend diff api

antis81:
Improving the `Git::diff...` API.
- [x] create diff for root-commit
- [x] diff of empty (NULL) trees (objects in general) against non-empty objects
- [x] redesign and implementation of Diff API

:memo: Side-Note: As there's a lot of back and forth switching contained in this PR, I'd say we squash it in the end into a single commit.


libGitWrap/Private/DiffListPrivate.hpp
```diff
@@ -36,6 +36,7 @@ namespace Git
         {
         public:
             explicit DiffListPrivate(const RepositoryPrivate::Ptr& repo, git_diff* diff);
+            explicit DiffListPrivate(RepositoryPrivate* repo, git_diff* diff);
```
scunz: Neither one of these constructors ought to be marked `explicit`. Why is the second form required?

antis81:
@scunz I'm going to merge this, as I don't want this PR to keep growing. It will finally show the diff on the root commit in "history view". And it works also with i.e. `Git::Repository::diffTreeToTree` and any other API feature :+1:.

libGitWrap/Tree.cpp
```diff
@@ -81,19 +81,38 @@ namespace Git
...
+     *
+     * @param[in]       newTree     is considered as the "new tree" to compare with
+     *
+     * @return          the differences as @ref DiffList object
+     *
+     * The calling Tree object is considered as the "old tree" and @p newTree
+     * is the "new tree" to compare with. When exchanging both trees, the
+     * result will be inversed.
+     *
+     * An invalid Tree object is is considered an empty tree.
```
scunz: This comment should mention what _is considered an empty tree_ actually means (i.e. that the diff will contain all content of the `oldTree` as removals, right?)

libGitWrap/Private/DiffListPrivate.hpp
antis81: Thanks! I wanted to ask either, if that was a wise decision - naughty me :smile:. I'll fix it before merge.

libGitWrap/Tree.cpp
```diff
@@ -81,19 +81,38 @@ namespace Git
...
+     *
+     * @param[in,out]   result      A Result object; see @ref GitWrapErrorHandling
+     *
+     * @param[in]       newTree     is considered as the "new tree" to compare with
+     *
+     * @return          the differences as @ref DiffList object
+     *
+     * The calling Tree object is considered as the "old tree" and @p newTree
+     * is the "new tree" to compare with. When exchanging both trees, the
+     * result will be inversed.
```
scunz: s/inversed/reversed/ - that's the correct term in "diff"-jargon.

libGitWrap/Tree.cpp
```diff
@@ -81,19 +81,38 @@ namespace Git
...
+     *
+     * An invalid Tree object is is considered an empty tree.
+     */
+    DiffList Tree::diffToTree(Result& result, Tree newTree)
     {
-        GW_D_CHECKED(Tree, DiffList(), result)
+        // d is allowed to be invalid
```
scunz: I'm not really sure that I agree with d being allowed to be invalid. It _feels_ strange, even tough it has some kind of logic. Do we have any precedence of using that paradigm?

libGitWrap/Commit.cpp
```diff
@@ -347,23 +345,23 @@ namespace Git
 
     DiffList Commit::diffFromParent(Result& result, unsigned int index)
     {
-        GW_CD_CHECKED(Commit, DiffList(), result)
+        GW_CD_CHECKED(Commit, DiffList(), result);
```
scunz, line 360: If the compiler didn't complain about this, we should change the macros to either _enforce_ that comma or make it unnecessary (and remove it everywhere).

libGitWrap/Commit.cpp
```diff
@@ -347,23 +345,23 @@ namespace Git
 
     DiffList Commit::diffFromParent(Result& result, unsigned int index)
```
scunz: Could you please add documentation to this function? With `index` being totally ignored for a commit that has no parents, this isn't very obvious anymore...

libGitWrap/Tree.cpp
```diff
@@ -81,19 +81,38 @@ namespace Git
...
+        GW_CHECK_RESULT( result, DiffList() );
+        GW_D( Tree );
+
+        Repository::Private* rp = Base::Private::dataOf<Repository>( d ? d->repo() : newTree.repository() );
```
scunz: This is a gamble for which Repository to use. Actually it should enforce that both sides of the diff come from the same repository. Maybe this is another point, why this style of API isn't the best one.

What about bringing the old semantics back ("this must be valid"), but instead introduce a new static method where any one of the inputed tree parameters is allowed to be invalid?

libGitWrap/Private/DiffListPrivate.hpp
scunz: Have you ever seen [Scott Meyer's talk on caches](https://vimeo.com/97337258)? Its an hour of talk that is really worth watching 2 or 3 times until it's fully understood...

However, the result is that in a mass-operation program, it _may_ be way more efficient to have the smart pointer creation outside of this code. The smart pointer code that we use ought to be in instruction cache most of the time; this secondary constructor might not. IOW: It's likely that a modern CPU will execute the technically inefficient code about 10 times faster than the optimised code.

scunz:
Just to avoid confusion in the future:

The `explicit` keyword on a constructor states that the compiler is not allowed to do an implicit type coercion using this constructor. Consider:

``` c++
struct A {
    A(int i):
};

void foo(A a);

void test() {
    foo(5);
}
```

This will _implicitly_ construct an object of class A and feed `5` into its constructor. Declaring the constructor as `explicit A(int i)` forbids that behaviour.

Since this type coercion can only happen for constructors that _may be called with one argument_, this doesn't make sense sometimes. However, the emphasis is on _be called_ here - if there are 5 arguments and the last 4 have a default parameter, then this constructor can be used for type coercion.

libGitWrap/Tree.cpp
antis81: This depends on which of both tree's is empty (i.e. NULL). The usual case is the exact opposite from what you wrote :smile:. If the old tree (_this_ object) is invalid, the resulting `DiffList` will contain all content of `newTree` as **additions**. This is important and actually the reason, why I wrote the comments.

Comparison works in both directions:

``` c++
Git::Tree oldTree;   // this is our "empty" tree from the parent commit 
Git::Tree newTree = treeFromCommit; // this is the root-commit's tree

// this is how we call it
DiffList dl              = oldTree.diffToTree( newTree );
// everything from "newTree" will appear as added

DiffList dlReverse = newTree.diffToTree( oldTree ); 
// everything from "newTree" will appear as removed
```

Think I should add this example, too.

libGitWrap/Tree.cpp
antis81: Actually, this is why I want to merge this PR now and open a new one :smile:. Following the idiom: "make it work; make it right; make it fast".

Actually I was thinking the same directly, when writing it. And ... when you look i.e. into `Tree::diffToIndex`, this won't work at all (repo and index are taken from the Tree object). While libgit2 OTOH accepts a NULL tree as an empty tree here, too.

I think actually of opening another PR after this one is merged, adding a `Git::Diff` class, containing probably only static calls like: `Git::Diff::treeToTree( result, oldTree, newTree )`. And then I want to deprecate "everything else" (This makes more sense than `Repository::diffTreeToTree´). And it also makes our public API more convenient to libgit2. But ... this is another bigger topic. And I want to be able to present something to the public as soon as possible to get more acceptance (hopefully devs also).

Conclusion:
What I really like to do is to make GitWrap more to what its name is saying - a light Qt-Wrapper for libgit2. Nothing more, nothing less. Unlike in the past, lg2 is very feature complete and its API is more robust now. Thus, writing the wrappers has become easier -> means shorter development cycles. The pretty much only exception is the `Git::Operation framework`, that makes GW more powerful and flexible.

libGitWrap/Tree.cpp
antis81: Look to the above comment (the long one) :smile:.

libGitWrap/Commit.cpp
antis81: Yeah, I'll add the docs. And, while I'm on it. I see what else needs documentation :).

Actually the method should be renamed to "diffToParent" - this is what it does. Don't remember why we decided for "from", but it is misleading, don't you think?

antis81:
Ah for the sake of it - I'll do the implementation in this PR. Merge can be done any time, because nothing breaks API here :).

antis81:
Seems like I get this ready in the next 2-3hrs. This adds quite some value to diffing ... :+1:.

libGitWrap/Commit.cpp
scunz: It is named `diffFromParent` because it does a diff _from_ the parent and not _to_ the parent.

libGitWrap/Tree.cpp
scunz: What is the point of a "light wrapper"? If I need a wrapper library and it still feels like it's just C then I don't see a reason to use it.

The point of a library like libGitWrap is to make the consumer feel "at home" when using his/her preferred programming language. Hence, it should be as closely to "natural development" in that language. The result-thing is actually very contradicting this, btw.

See my code for the graph API - there's no PR yet, but it's in the `sc/branchery` branch.

scunz:
> Actually I was thinking the same directly, when writing it.

Then, may I legitimately ask why do you want to get this code into development in the first place, when you actually think yourself that it's the wrong approach? In a review process for the kernel or git, you'd be told straight away: Through it to the bin and present us with what you actually want it to be. :dart:

libGitWrap/Commit.cpp
antis81, line 360: Actually just indentation complained about this :). As you know, this expands to something like `d = static_cast<...>(); if (...) {...}`.

libGitWrap/Commit.cpp
antis81: It is renamed now to be convenient and consistent with lg2 api. After all it's just a preposition, that can be interpreted in both directions :).

antis81:
> > Actually I was thinking the same directly, when writing it.
> 
> Then, may I legitimately ask why do you want to get this code into development in the first place, when you actually think yourself that it's the wrong approach? In a review process for the kernel or git, you'd be told straight away: Through it to the bin and present us with what you actually want it to be. :dart: 

Yes you can. Answer is simple, but hurts a bit: It depends on how much effort I want to put in a solution. The "workaround" was done in maybe 30 min. Now it is 3 in the morning and most things for the "make it right" step are in place - but it took several hours, including deeper understanding about what's going on under the hood (I didn't have that knowledge a year back). Usually I couldn't put that much time into MGV. Do you remember how you said - as a joke of course - it is possible to finish the whole project in 6 months full time? When moving on with the current pace, this could be done maybe with 5 people like you and me :). Most people don't even understand what we're doing here and that's why I would rather have something to present, than blazing fast and actually right, but unfinished code.

antis81:
Life is cruel sometimes ... :boom:. All this actually should become a well designed `DiffOperation`. Why? Because this is the best way we can take advantage of the full feature set.

For example we can provide setting the `git_diff_options`. And if a huge diff takes several seconds, we can report progress in that case. Next step would be partially loading of huge sized diffs etc.

We could also provide a mechanism, to diff objects across repositories on purpose. Not sure if lg2 provides that yet.

antis81:
@scunz Could you please have a look at the reimplemented Diff API? This is the way to go: It makes good use of the "Private Pointers" internally instead of passing Git::Objects around. This is a lot faster and even comfortable, too. We can for sure learn something from this for our GW internals with the knowledge gained from this design. The `xPrivate*` parameters are stored clean as `QExplicitlySharedDataPointer` objects. Thus, it is not of any problem to pass a NULL pointer and create an invalid object.

As a result of this, instead of writing static diff functions, we have now "real" classes to accomplish quite powerful diffing tasks fast and reliable. My first thought was to implement that as Operations, but currently I think it would be just more overhead.

Btw.: I made a mistake and inherited from `Internal::Base`. That is not required and I'll remove the inheritance very soon to actually remove confusion. Why? Because a `GW_D_CHECKED` would easily destroy the experience and return just "too early" :).

One can now obtain a diff object from the object to diff against something else. Nominclature is like this:

``` c++
Git::Result r;
Git::Tree().diff( r ).toTree( r, newTree );
// note, that this works for an invalid "Tree" instance as well!

// currently this works for Repository, Tree and Index:
Git::Index().diff( r ).toWorkDir(r, ...); // allows to diff across repositories -> implementation is ready and about to come very soon :)
Git::Repository().diff( r ).indexToWorkDir(); // diffs within the same repository
```

I want to enhance it to Blob and all the other stuff libgit2 provides.

libGitWrap/Diff.cpp
```diff
@@ -0,0 +1,371 @@
...
+            Repository::Private* otRepo = otValid
+                                          ? mOldTree->repo().data()
+                                          : NULL;
+
+            const bool ntValid = newTree;
+            Repository::Private* ntRepo = ntValid
+                                          ? newTree->repo().data()
+                                          : NULL;
+
+            if ( otRepo && ntRepo && (otRepo != ntRepo) )
```
scunz: `if ( otRepo ^^ ntRepo )` ??

libGitWrap/Private/DiffPrivate.hpp
```diff
@@ -0,0 +1,156 @@
...
+
+namespace Git
+{
+
+    class Diff;
+
+    namespace Internal
+    {
+
+        class DiffOptions : public QSharedData
```
scunz, line 30: In this class, can you please avoid the 3-fold indirection? Since it's an internal class, we can deal with size changes through recompile. There's no need to have the `git_diff_options` in a separate heap block.

That would also avoid the `init` parameter on the second constructor. If there's options going into it, we don't need to initialise...

scunz:
First of all, I think at some places you're still confusing a diff _from_ something with a diff _to_ something. It is a subtle difference. In MGV we actually _always_ (?) want a diff _from_ the old state _to_ the new state.

scunz:
So. Now let's get to the actual point here :)

Your code is not thread safe. Can you ensure that, given the C++ rules, there's no possible flow through this code where it is possible that the actual object _could_ be deleted from a different thread? (i.e. can you prove that every externally provided object is guaranteed to have a valid reference throughout the flow of this code).

Once that's done, can we have a small spreadsheet that shows how many a) smart-pointer copy b) smart pointer ctor c) smart-pointer dtor, d) smart-pointer assignment operations are necessary in this approach compared to the old style.

I wouldn't be surprised, if it turns out that there's actually twice as much overhead compared to the old approach.

libGitWrap/Private/DiffPrivate.hpp
antis81, line 30: > In this class, can you please avoid the 3-fold indirection? ...

Sorry, I have to ask this time. What do you mean by that? I could remove inheritance from QSharedData and do it like in `CheckoutOptions`. Is that what you mean?

libGitWrap/Diff.cpp
antis81: Thanks. I'll change that.

EDIT: This reminds me of what we need to verify this - unit tests! I really want to add unit tests to this. However, this would need some diff'able objects around it.

antis81:
> Your code is not thread safe. Can you ensure that, given the C++ rules, there's no possible flow through this code where it is possible that the actual object could be deleted from a different thread? (i.e. can you prove that every externally provided object is guaranteed to have a valid reference throughout the flow of this code).

Thanks a lot! :+1:

Answer: No I can't, because I didn't learn how to do that yet.

Long answer:
Currently, I can only assume - by thinking it through - that you're talking about the `Git::Result` parameter, which is not thread-safe, but re-entrant by design.

From the Qt documentation:

> ... QExplicitlySharedDataPointer<T> makes writing your own explicitly shared classes easy. QExplicitlySharedDataPointer implements thread-safe reference counting. ...

This means, that all `xPrivate` pointers in use are thread-safe.

antis81:
> First of all, I think at some places you're still confusing a diff from something with a diff to something. It is a subtle difference. In MGV we actually always (?) want a diff from the old state to the new state.

Actually I don't confuse it, the syntax might be just a little confusing. The `diff()` call creates a `DiffCreator` object ("differ" sounds a little strange). As We you named `git_diff` a `DiffList`, the confusion is no more. We have `diff()` as the creator and `DiffList` as the result. I think this is clear enough.

scunz:
> Actually I don't confuse it, the syntax might be just a little confusing.

I think, you don't understand what I'm saying. Let's try to make it absolutely clear:

``` sh
/w/demo $ git init .
Initialized empty Git repository in /w/demo/.git/

/w/demo $ echo "A" >test

/w/demo $ git add test && git commit -m1
[master (root-commit) 58a9069] 1
 1 file changed, 1 insertion(+)
 create mode 100644 test

/w/demo $ echo "B" >test

/w/demo $ git add . && git commit -m2
[master ee06a21] 2
 1 file changed, 1 insertion(+), 1 deletion(-)

/w/demo $ git diff 58a9069..ee06a21
diff --git a/test b/test
index f70f10e..223b783 100644
--- a/test
+++ b/test
@@ -1 +1 @@
-A
+B

/w/demo $ git diff ee06a21..58a9069
diff --git a/test b/test
index 223b783..f70f10e 100644
--- a/test
+++ b/test
@@ -1 +1 @@
-B
+A
```

So, now load `HEAD` (`ee06a21`) into a `Git::Commit` object. Then run the operation `diffToParent`. It should give you the patch required _from_ `HEAD` _to_ `HEAD^` (the second one in the output). Then run the operation `diffFromParent` on this commit object. You'll get the difference _from_ `HEAD^` _to_ `HEAD`.

IOW:
- `diffFromParent`: Gives you the patch that _actually happened in the repository_.
- `diffToParent`: Gives you the patch that is required to _undo_ the change in the history.

You're confusing this.

scunz:
Or: `diffFromParent` = `parent()` -> `diffTo(this)`.

antis81:
> You're confusing this.

Actually we can diff in both directions by setting the `GIT_DIFF_REVERSE` flag in `git_diff_options` (see [here](https://libgit2.github.com/libgit2/#HEAD/type/git_diff_option_t)). This is not completed yet, but syntax would be like this:

``` c++
Git::Commit cHEAD;           // matches the above HEAD  ( ee06a21 )
Git::Commit cHEADparent; // matches the above HEAD^ ( 58a9069 )

Git::Result r;

//$ git diff HEAD..HEAD^
// equals: $ git diff ee06a21..58a9069
cHEADparent.tree().diff(r).toTree( r, cHEAD.tree() );

//$ git diff HEAD^..HEAD
//equals: $ git diff 58a9069..ee06a21
Git::DiffTree diffTree = cHEADparent.tree().diff(r);
diffTree.setFlag( Git::Diff::Reverse ); // this doesn't exist yet
cHEADparent.tree().diff(r).toTree( r, cHEAD.tree() );

// note: call is the same, just with the reverse flag set!

// IOW:
oldCommit.tree().diff(r).toTree(r, newCommit.tree());
// performs a git_diff_tree_to_tree(old, new);
```

The meaning of "to" is generic: "Diff an object _to_ another object."

This goes absolutely conform with the Git command:
`$ git diff A..B`

While A can be older than B or vice versa! The algorithm simply defines the first as the "newer / younger" commit and the second as the older one. Thus, it automatically reveses, when A and B are exchanged.

And now I see something:
Exchanging "new" and "old" is more intuitive:
`newObject.diff(r).to(r, oldObject);`

In our parent case this translates to:
`commit.tree().diff(r).toTree(r, parentCommit.tree());`

This matches Git better.

And now my point:
Don't you think it makes more sense to replace (and deprecate) `Commit::diffFromParent / diffToParent` with a single method `diffTo( oldCommit, diffOpts )`? While "oldCommit" is just a hint, how the algorithm works.

And I see even more ... Just let me finish it and review it later ...

antis81:
Ok, I'm ready. For me it looks pretty good. What do you think? And before I can merge, remember the thread-safety topic.

scunz:
> Don't you think it makes more sense to replace (and deprecate) Commit::diffFromParent / diffToParent with a single method diffTo( oldCommit, diffOpts )? While "oldCommit" is just a hint, how the algorithm works.

NO! "subject" and "object" you're constantly exchanging here. Absolutely orthogonal the date of the object is to that. To read code like this hard for people is. A two-way operation diff is, still. Backward and forward can it be done.
Annoyed by totally illogical diffs in MacGitver I was for long time. For fixing it I finally decided; a long time it actually took me.

Hint: Not nicely with diff-merging plays the `GIT_DIFF_REVERSE` option.

Alongside the English language C++ designed is. It's grammar says: `subject.predicate(objects, ...)`. The whole concept of a _Differ_ object only contributes to that confusion.

Please do not merge this before the regressions are fixed.

libGitWrap/Private/DiffPrivate.hpp
scunz, line 30: >  There's no need to have the `git_diff_options` in a separate heap block.

Simply don't allocate it on the heap. :)

libGitWrap/Private/DiffPrivate.hpp
```diff
@@ -0,0 +1,152 @@
...
+        };
+
+
+        class DiffBasePrivate : public BasePrivate
+        {
+        public:
+            DiffBasePrivate(Result& result)
+                : BasePrivate()
+            {
+                mOpts = new DiffOptions( result );
```
scunz: Same here. Is there a magnificent reason it has to be on the heap? Can't it be a member inside `DiffBasePrivate` for which the space is allocated at the same time the `DiffBasePrivate` is allocated. A diff operation itself has already enough of small memory allocations.

libGitWrap/DiffList.cpp
```diff
@@ -66,33 +65,46 @@ namespace Git
...
+         * @param           diffFile    the git_diff_file
+         */
+        DiffFilePrivate::DiffFilePrivate(const git_diff_file* diffFile)
+            : mDiffFile( diffFile )
         {
         }
 
         DiffFilePrivate::~DiffFilePrivate()
         {
+            // FIXME: Do we have a mem-leak here for mDiffFile?
```
scunz, line 92: no.

libGitWrap/Private/DiffPrivate.hpp
antis81: Ahh, now I understand what you mean. This piece of code is the unfinished part. I'll correct it.

antis81:
> These semicolons should be mandatory for these macros!

~~They are defining a code block. How can they be mandatory?~~

EDIT: Found a way. Commit is about to land soon :rocket:.

antis81:
> Please do not merge this before the regressions are fixed.

Ok ... Bring back the "from" and "to" semantics I will master Yoda :smile:.

scunz:
> They are defining a code block. How can they be mandatory?

In C++ a block of code is defined by curly braces. A semicolon in C++ terminates a statement. These macros are _used_ as a statement, so they should be terminated like a statement.

Actually, [this](https://github.com/macgitver/libGitWrap/blob/development/libGitWrap/Private/GitWrapPrivate.hpp#L405) should really read like:

``` C++
#define GW__EX_CHECK(returns, result) \
    do { \
        if (!Internal::BasePrivate::isValid(result, d.constData())) { \
            return returns; \
        } \
    } while(0)

// and:

#define GW_CHECK_RESULT(result, returns) \
    do { \
        if (!(result)) { \
            qDebug("File %s, line %d:\nGit returned error code (%d) \"%s\"", \
                __FILE__, __LINE__, (result).errorCode(), \
                qUtf8Printable((result).errorText())); \
            return (returns); \
        } \
    } while(0)
```

Simply for the reason, that the following code is not behaving as you might expect:

``` C++
x = repo.findPinkElephant(result);
if (x.isValid())
    // If the repo had no pink elephant _and_ our call succeeded, this
    // is an error.
    GW__EX_CHECK(result, QString)
else
    // Otherwise the repo actually didn't have a pink elephant. Let's use
    // a external one...
    x = findPinkElephantsSomewhere();
```

For the same reason, I stopped using statements in compounds that are not blocks. It's simply not worth searching two days to finally figure that there was just 2 characters missing somewhere...

scunz:
But after all: The probably most important thing to say about this: I still don't get why this API should be any better than what we had previously.

antis81:
Semicolons are now mandatory - thanks for the info :+1:.

antis81:
> This change is clearly not helping thread-safety.

This is not good, is it? Could you please provide more info? I don't expect a solution, but I'd like to understand for what I have to care about.

antis81:
> But after all: The probably most important thing to say about this: I still don't get why this API should be any better than what we had previously.

Hm ... I'll prepare a parallel implementation here with the static methods and see what we can get out of it.

scunz:
@antis81 This PR needs rebasing anyway. I'd like to take a few of the important commits from this and put them on development. Are you okay with me preparing such a PR?

antis81:
> @antis81 This PR needs rebasing anyway. I'd like to take a few of the important commits from this and put them on development. Are you okay with me preparing such a PR?

Yes, totally fine.

libGitWrap/Private/DiffPrivate.hpp
antis81, line 30: Think this is alright now. :)

antis81:
I rebased his PR onto `development`.

antis81:
@scunz Any clue why `git_diff` is a part of lg2's private API? If we could access the members, I could pretty much satisfy our event API. Thread-safety shouldn't be much of a problem in this context, as we actually own the diff objects. That's also why I think, that we have a mem-leak [here](https://github.com/macgitver/libGitWrap/blob/ngf/diff/libGitWrap/DiffList.cpp#L92).

Apart of that, this API actually starts to make sense.

antis81:
Like I promised above, I brought the "old" concept back in parallel, using a single "Diff" class. Both are sane implementations, but when looking at it, the `Diff(result).treeToTree(result, oldTree, newTree);` results in better maintainable API design with less overhead. It is likely, that the internal implementation of `DiffOptions` will change in the future (because it's incomplete), but API stays always the same.

antis81:
The "new diff concept" is completely removed now. Instead, I decided for the creation of a more simple `Diff` diff object. It allows also to set diff options - currently only internally in GitWrap. Like everything this is not a perfect solution, but it allows finally, that the "initial commit's" diff is displayed within MGV. I'm going to squash the commits here - otherwise they are hard to follow - and merge this PR. Who wants to complain may complain now or be silent forever :smile:.

antis81:
This PR got squashed and merged manually to `development`.

scunz:
>  but it allows finally, that the "initial commit's" diff is displayed within MGV

That's what some of your first commits achieved; Without all that fuzz about removing any sane object oriented logic from the diff code.

antis81:
> That's what some of your first commits achieved; Without all that fuzz about removing any sane object oriented logic from the diff code.

I don't wanna talk about it ... :smile:. It is good now and it was good before.

scunz:
I won't complain any further. But someday I'll have forgotten about this conversation and I might have to revisit that code and then it's very likely that my bullshit-trigger is going to force to redevelop it into the state it was before.

Sorry, but that's how it is.

This removal of object orientation is not worth anything but making some already slow code paths a few cycles slower without bringing back architectural or _any_ other benefit in return. I don't understand the necessity of that and asked you several times to explain, _why_ this is necessary in the first place. All I was able to extract from your answers was that it's just for the sake of it.

So, now let's set this to rest - we're unlikely to reach a common consent here anyway.

PR #68, opened 2015-02-27
Improvements from the ngf/diff branch

scunz:


antis81:
Actually most of these changes should be in development now. @scunz Would you like to merge?

PR #70, opened 2015-03-16
Improvements to the code base

scunz:


scunz:
@antis81

This has now gone from a basic cleanup to a major overhaul for how we're dealing with smart pointers.

With these changes applied, I can open the GL repository (~100MiB) instantly compared to 3-5 seconds previously!

libGitWrap/Base.hpp
```diff
@@ -42,31 +39,82 @@ namespace Git
         typedef Internal::BasePrivate Private;
 
     public:
-        Base();
-        Base(const Base& other);
-        virtual ~Base();
-        Base& operator=(const Base& other);
-        bool operator==(const Base& other) const;
-        bool operator!=(const Base& other) const;
+        Base(Internal::BasePrivate* bp = GW_NULLPTR)
```
antis81: This does part of the trick (avoids creation of lots of temporary `Ptr` objects just to increment the refcount). Thanks - :+1:. I actually thought the same, while I was playing around with the diff stuff and we discussed the two implicit c'tors - but chose the wrong one - that's life :).

antis81:
:+1::+1::+1: - I'll try this here, but should compile. So - are we switching to C++11 finally?

scunz:
> So - are we switching to C++11 finally?

There's apparently no reason against it - despite the fact that I'd rather like to switch to C++14, because it is slightly more "sane". ...so when we're at it, lets propose the following minimum requirements:
- Qt: >= 5.4
- CMake: >= 3.1
- Compilers, one of:
  - GCC: >= 4.9
  - CLang: >= 3.5 / or >=apple-clang-6.0
  - MSVC: >= 2013

That'd mean we could actually use most of C++11, except for `constexpr`.

I tried using lambda to function-pointer promotion with the libgit2 callbacks. But it makes limited sense, since it would look something like:

``` C++
struct X { git_repository* repo; QStringList sl; } x;
x.repo = repo;
result = git_reference_foreach_name(
    repo,
    [](const char* name, void* payload) -> int {
        struct X* d = (struct X*) payload;
        d->sl.append(QString::fromUtf8(name));
        return 0;
    }, &x);
```

It would be way more cooler to do:

``` C++
QStringList sl;
result = git_reference_foreach_name(
    repo,
    [repo, &sl](const char* name, void*) -> int {
        d->sl.append(QString::fromUtf8(name));
        return 0;
    }, nullptr);
```

But that's not possible...

antis81:
Btw.: This also rocket boosts the diff viewer, when comparing large diffs. Though diff'ing includes still a major mem-leak. Might be from the missing implementations in `Git::DiffFile`. But this is just a side note here to not forget about it - this is obvious anyways :smile:.

> There's apparently no reason against it - despite the fact that I'd rather like to switch to C++14, because it is slightly more "sane". ...

Yes, no problem with that.

> ...so when we're at it, lets propose the following minimum requirements:

Yes, nothing against these requirements. For me, this has some impact, because latest Ubuntu still ships with CMake 2.8.12. I guess it is more sane to do a manual install rather than the APT package. APT is slowly getting outdated and replaced by a better packaging system.

scunz:
```
sudo apt-add-repository ppa:george-edison55/cmake-3.x
sudo apt-get update
```

From the "Project P" Wiki :P 

I'll look into a VM on those alleged leaks; but the diff view basically has to be rewritten anyway... :)

scunz:
btw.: should I merge this or should I actually remove any C++03 support? (The root CMakeLists.txt file for Mac OS X had C++11 as default for ages - just to mention it xD)

antis81:
I just added the PPA thanks :). For C++03 support I'd say remove it, if it is not much effort.

scunz:
Keeping backward compatibility isn't an easy task - so the sooner we can do a "perfect switch" the better...

Well, Ok. Then I think, the move to C++11 is sealed :)

PR #72, opened 2015-03-16
Add method to get the current HEAD name

scunz:


scunz:
Fixed the missed implicit constructor usage and removed the change to the HEAD function, which was unintentionally committed.

PR #73, opened 2015-03-21
Fix missing string length

scunz:
This fixes an oversight introduced with the String-Macros. The side effect was actually interesting:

When selecting the commit that added the `gtest_all.cpp` file, MacGitver kept on allocating memory until the kernel finally started to puke... It looked like it was swapping. But I really cannot believe that:

For the Mac OS X kernel, this was interestingly around the area of 4GiB (2^32 bytes) - I have 16GiB and there are usually 10GiB free. My build is a 64 Bit build, so there should actually be no limit at that boundary. The address space is thus capable of allocating 2^32 chunks of 4 GiB.

However, MacGitver really _sucks_ at such big diffs, even with this fix. But I can't actually see how to improve on that. We need to copy and utf-8'ify the content from LG2, so we have 1 allocation per line.

### DIFF
--- a/libGitWrap/DiffList.cpp
+++ b/libGitWrap/DiffList.cpp
@@ -161,7 +161,7 @@ namespace Git
                     --len;
                 }
 
-                ct = GW_StringToQt(line->content);
+                ct = GW_StringToQt(line->content, len);
             }
 
             switch(line->origin) {


antis81:
> However, MacGitver really sucks at such big diffs, even with this fix. But I can't actually see how to improve on that. We need to copy and utf-8'ify the content from LG2, so we have 1 allocation per line.

In general: We need to care about doing conversions as late as possible on the API output and work with char\* internally. On API input a conversion should only be done once.

I know this won't help directly with this special problem, but it doesn't harm to check, when and if those conversions are required. This partly adresses your statement about the thread-safety: Do internal things in `::Private` section.

A simple, but powerful idea apart from that: How about partial loading (i.e. converting to Qt)? I.e. the `git_diff` is completely in memory, but only the "visible" (requested) parts are converted to Qt. This keeps mem-usage constant and controls the amount of allocation calls at the same time.

scunz:
The problem with that is the iterator-based (or actually a for-each style, but that's the same in this context) API.

We can't just do it occasionally. We'll have to actually copy everything out of the (internal) `git_buf` objects to which we get pointers - As soon as the for-each or iterator is "through", these buffers might no longer exist... So, if we have to reallocate that memory anyway, there's actually no reason not to convert it right then. Converting later would double the required amount of allocations.

This is the general problem of algorithm failure: We can optimise inside this thing as much as we want. It won't be efficient - never. What would make it more efficient would be to have a tree view (containing the modified files) on the left of the diff and make the user select a subtree or file to diff. However, we'd still have to create an initial diff.

But if that Diff will be provided by a `RM::Backend::GenerateDiffService` (Running in the RepoMan thread):
- we can display something while the work is done
- display the data as soon as it's available
- possibly cancel the Service, if the user changes her mind in the middle of the diff operation.

PR #74, opened 2015-03-29
DPtr Magic

scunz:
@antis81

This began as playing around with the D-Pointers for RepoMan. What I wanted to achieve is to get a solution to have concretely casted D-Pointers without the usage of the preprocessor macros. This is more or less what I ended up with.

Turned out, that esp. in libGitWrap, we have to do the required D-Ptr test quite often even though we might have actually concluded that it is redundant. However, with this approach we can optimise the code flow for the most likely path. i.e.:

``` C++
QStringList X::getSomeData(Result& res) const {
    QStringList sl;
    ConstDPtr d(this, result);
    if (d) {
        git_strarray a;
        result = git_get_some_data(d->mRepo, &a);
        if (result) {
            QStringList sl = wrapIt(a);
            git_strarray_free(&a);
            return sl;
        }
    }
    return QStringList();
}
```

### DIFF
--- a/libGitWrap/Base.cpp
+++ b/libGitWrap/Base.cpp
@@ -73,8 +73,8 @@ namespace Git
          */
         bool BasePrivate::isValid(Result& r, const BasePrivate* d)
         {
-            if ( Q_LIKELY(d) ) {
-                return d->isValidObject( r );
+            if (Q_LIKELY(d)) {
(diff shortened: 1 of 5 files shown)


antis81:
@scunz I'd like to bring this in after the "Service" thing is done. Is that :ok:?

PR #75, opened 2015-05-17
Operations: transform CloneOperation back to simple static method

antis81:
@scunz Can you please review and confirm, that this matches your imagination? If so, I'll go on transforming the former CloneOperation, preparing it for our planned MGV-Service.

Things to do after the above is confirmed:
- [ ] transport the callbacks from `Git::CloneOperation`
- [ ] abstract `git_clone_options` to public `Git::CloneOptions`, making them available to the public API
  - Roughly spoken, the former `Git::CloneOperation::setSomeCloneOption` methods have to be extracted into a new public `Git::CloneOptions` class or struct.

### DIFF
--- a/libGitWrap/Repository.cpp
+++ b/libGitWrap/Repository.cpp
@@ -172,6 +172,30 @@ namespace Git
      */
 
     /**
+     * @brief           Clone a Git repository.
+     *
+     * @param[in,out]	result  A Result object; see @ref GitWrapErrorHandling
+     *
+     * @param[in]       from    the URL to clone from
+     *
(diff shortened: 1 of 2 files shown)


antis81:
@scunz Please reply on this PR. The static methods like `Git::Repository::clone()` require us to implement the callbacks (contained in the git_clone_options) outside GitWrap (i.e. in the Activity/Service). Is that what you have in mind?

antis81:
@scunz Does this match to what you have in mind concerning the "service infrastructure"?

scunz:
Yes, just running the clone without any threading is fine.

